PR #5, opened 2020-07-22
add groupable field

rleiwang-traceableai:
With groupable field, UI will render attribute in group-by widget iff groupable == true

### DIFF
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
@@ -47,6 +47,9 @@ public class AttributeMetadataModel implements Document {
   @JsonProperty(value = "tenant_id")
   private String tenantId;
 
+  /** if true, UI will render the attribute in group-by widget */
+  private boolean groupable;
+
   private List<AggregateFunction> supportedAggregations = Collections.emptyList();
   private boolean onlyAggregationsAllowed;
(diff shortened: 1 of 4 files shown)


aaron-steinfeld:
How does this get defaulted?

rleiwang-traceableai:
the default value is false, at https://github.com/hypertrace/attribute-service/pull/5/files#diff-ea085817933ca044a95fd50b543cb420R51

aaron-steinfeld:
> the default value is false, at https://github.com/hypertrace/attribute-service/pull/5/files#diff-ea085817933ca044a95fd50b543cb420R51
1. Please remove the comments about the UI - they're irrelevant to attribute service. This is a general setting that can be used throughout the platform (we should enforce this in query at some point later)
2. I disagree with the default - we should maintain the current behavior, which is that any string typed attribute is by default groupable. Config should have to opt out, rather than opt in.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -47,8 +47,7 @@
   @JsonProperty(value = "tenant_id")
   private String tenantId;
 
-  /** if true, UI will render the attribute in group-by widget */
-  private boolean groupable;
+  private boolean groupable = true;
```
aaron-steinfeld, line 50: Not all attributes are groupable though - we just want strings right now. I don't think we get away from logic here. How do the other fields get their defaults set? The default aggregations for example are set based on the attribute type.

PR #11, opened 2020-08-04
set groupable to true for all TYPE_STRING fields

rleiwang-traceableai:


helm/configs/1_event_attributes.conf
```diff
@@ -12,6 +12,7 @@ commands = [
               fqn: Span.id,
               key: id,
               value_kind: TYPE_STRING,
+              groupable: true,
```
buchi-busireddy, line 15: I thought `true` is the default value for all string attributes.

aaron-steinfeld: Because of the config bootstrapper set up, all attribute definitions get pumped through proto serde so appear with the proto defaults. For booleans, that'd be false - we have to set them explicitly.

PR #16, opened 2020-08-19
feat: add caching attribute client

aaron-steinfeld:


caching-attribute-service-client/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
+  `java-library`
+  jacoco
+  id("org.hypertrace.jacoco-report-plugin")
```
avinashkolluru, line 4: publish?

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/DefaultCachingAttributeClient.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Override
+  public Single<List<AttributeMetadata>> getAll() {
+    return this.getOrInvalidate(AttributeCacheContextKey.forCurrentContext())
+        .map(table -> List.copyOf(table.values()));
+  }
+
+  private Single<Table<String, String, AttributeMetadata>> loadTable(AttributeCacheContextKey key) {
+    return key.<AttributeMetadata>streamInContext(
```
avinashkolluru, line 61: Took me some time to wrap my head around this.
It is this vs exposing a getter for the RequestContext in AttributeCacheContextKey and directly calling `GrpcClientRequestContextUtil.executeWithHeadersContext` here.

caching-attribute-service-client/build.gradle.kts
aaron-steinfeld: Not being used yet, but might as well add it now I suppose.

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/DefaultCachingAttributeClient.java
aaron-steinfeld: Yeah, just delegating the responsibility of how to run something in the key's context out of this class. The cache key itself might not be the right place for it, but I don't think it's here either. Ideally, I'd probably create a context abstraction but I think given that we already have two context objects, that seems confusing. Another option would be to push this functionality into the platform context object or one of the associated utils - I think that's feasible for call and run, but I wouldn't want to introduce an Rx dependency there for stream.

A better answer might be to create a grpc stream utils package with the other ones so it can be opt in, as I've had to copy various forms of this pattern around a couple times now.

tim-mwangi:
@aaron-steinfeld will this replace the caches we have internally for attributes in gateway service for example?

aaron-steinfeld:
> @aaron-steinfeld will this replace the caches we have internally for attributes in gateway service for example?

Not as part of this effort, but that's the idea - gql too.

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/CachingAttributeClient.java
```diff
@@ -0,0 +1,120 @@
...
+     * Filter all attributes fetched and cached with the provided filter. Useful for limiting a
+     * client to use a particular scope, for example. Defaults to empty.
+     *
+     * @param attributeFilter
+     * @return
+     */
+    public Builder withAttributeFilter(
+        @Nonnull AttributeMetadataFilter attributeFilter) {
+      this.attributeFilter = attributeFilter;
+      return this;
```
tim-mwangi, line 117: Codecov for the methods in here

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/AttributeCacheContextKey.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hypertrace.core.grpcutils.client.GrpcClientRequestContextUtil;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+class AttributeCacheContextKey {
+  static AttributeCacheContextKey forCurrentContext() {
+    return forContext(RequestContext.CURRENT.get());
+  }
+
+  static AttributeCacheContextKey forContext(RequestContext context) {
+    assert !isNull(context) : "RequestContext must be set to use Attribute Client";
```
buchi-busireddy, line 19: assertions might not be enabled by default right? Should we enable them or rather convert this into a guava `Preconditions`?

aaron-steinfeld: I actually wrote this with requireNonNull and IJ suggested the change. Didn't realize assertions weren't enabled by default, will switch this back and fix that suggestion in my env.

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/DefaultCachingAttributeClient.java
```diff
@@ -0,0 +1,87 @@
...
+    this.cache =
+        CacheBuilder.newBuilder()
+            .maximumSize(maxCacheContexts)
+            .expireAfterWrite(cacheExpiration)
+            .build(CacheLoader.from(this::loadTable));
+  }
+
+  @Override
+  public Single<AttributeMetadata> get(String scope, String key) {
+    return this.getOrInvalidate(AttributeCacheContextKey.forCurrentContext())
```
buchi-busireddy, line 49: Curious, why are we invalidating the cache when the entry is missing? Let's say you're looking for a non-existing attribute, would every following call go over the network?

aaron-steinfeld: Yeah good discussion point. My thinking here was that a get call _should_ always work. If we are using a known key and scope and it is missing, is it because attributes have been recently updated and the cache is out of sync? Should we give an error without verifying that our state is correct? If it were invalid, that call would go over the network. Subsequent calls wouldn't unless they were for that invalid key. In other words, this behavior is basically saying only cache successes, make a remote call before marking failure so we know all failures are valid.

I could definitely be convinced to change this, I'm actually leaning in the other direction now as a non existent attribute might be requested many times in sequence.

buchi-busireddy: I think the case where we should definitely avoid querying over and over is, an invalid attribute is looked up and the call to server has succeeded but didn't return any result. That's a clear case that Cache can remember and return empty response next time when same invalid key is looked up (until cache expiry).

aaron-steinfeld: I'm so incredibly dumb. I removed this and spent like an hour trying to fix the test only to finally realize my test was right all along (saved the day!) and I had apparently forgotten why I originally wrote this and explained it completely wrong. So, ignore what I said above. What this is _actually_ doing is not invalidating when an entry is missing (that error is generated on line 51 and won't invalidate the cache), but retrieving the async table from the cache. If that async call had ended up in an error (for example, the downstream was unavailable) we don't want to cache that error state but retry. Once a call succeeds, even if it returned an empty table, we won't manually invalidate it.

aaron-steinfeld:
Dependent on https://github.com/hypertrace/java-grpc-utils/pull/5

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/CachingAttributeClient.java
aaron-steinfeld: Added for the ones that are reasonably verifiable.

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/DefaultCachingAttributeClient.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Override
+  public Single<List<AttributeMetadata>> getAll() {
+    return this.getOrInvalidate(AttributeCacheContextKey.forCurrentContext())
+        .map(table -> List.copyOf(table.values()));
+  }
+
+  private Single<Table<String, String, AttributeMetadata>> loadTable(AttributeCacheContextKey key) {
+    return key.getExecutionContext().<AttributeMetadata>stream(
```
avinashkolluru, line 61: 👍

avinashkolluru (approved):

PR #17, opened 2020-08-20
feat: add attribute definition api

aaron-steinfeld:


aaron-steinfeld:
Not married to this, but figured we could use this PR to discuss the metdata design.

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -137,3 +138,39 @@ message AttributeMetadataFilter {
...
+    string string_value = 1;
+    bool boolean_value = 2;
+    double float_value = 3;
+    int64 int_value = 4;
+  }
+}
+
+message ProjectionExpression {
+  ProjectionOperator operator = 1;
+  repeated Projection arguments = 2;
```
aaron-steinfeld, line 167: Do we want this to be variable arity? Could do most things with chaining, but more confusing

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -137,3 +138,39 @@ message AttributeMetadataFilter {
 message Empty {
 }
 
+message AttributeDefinition {
+  oneof value {
+    string span_path = 1;
```
aaron-steinfeld, line 143: not sure about this. The intent is to describe how to extract from span/trace data for simple raw/enriched attributes (since we don't want the caller to have to know how something is defined, this should encompass any attribute type). Could see it being repeated. Does it need to differentiate for trace? understand enriched vs raw?

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -137,3 +138,39 @@ message AttributeMetadataFilter {
...
+message ProjectionExpression {
+  ProjectionOperator operator = 1;
+  repeated Projection arguments = 2;
+}
+
+enum ProjectionOperator {
+  PROJECTION_OPERATOR_UNSET = 0;
+  PROJECTION_OPERATOR_CONCAT = 1;
+  PROJECTION_OPERATOR_HASH = 2;
+  PROJECTION_OPERATOR_DEFAULT = 3;
```
buchi-busireddy, line 174: what's the default operator?

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -137,3 +138,39 @@ message AttributeMetadataFilter {
...
+message AttributeDefinition {
+  oneof value {
+    string span_path = 1;
+    Projection projection = 2;
+  }
+}
+
+message Projection {
+  oneof value {
+    string attribute_key = 1;
```
buchi-busireddy, line 150: Is this like a foreign key into another attribute? When would we do that?

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
buchi-busireddy: I think we should keep it.

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -98,6 +98,7 @@ message AttributeMetadata {
   map<string, AttributeSourceMetadata> metadata = 13;
   string id = 14;
   bool groupable = 15;
+  Projection definition = 16;
```
buchi-busireddy, line 101: @aaron-steinfeld one fundamental question is, we defines this projection? Should it all come via the config bootstrapping? That might mean putting lot of span key paths in these configs.

aaron-steinfeld: That was the idea that I was thinking - unifying the definition of an attribute in one spot, spanning from the user's point of view all the way down to the input data. Without this, using attributes becomes different for raw attributes vs derived/mutable attributes and raw or enriched attributes additions/changes require code level changes. At least that's based on my current understanding.

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
aaron-steinfeld: It is. The simple case here would be something like an alias. But this is also used in how we recursively define something in terms of other attributes. Take the definition of an attribute like serviceId, assuming for sake of discussion we want it to be based solely on service name:
```
 definition = {
   expression = {
     operator = HASH
     arguments = [{
       attribute_key = "serviceName"
     }]
   }
 }
```

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
aaron-steinfeld: the idea here is something like fallback/default value. Use the value of X attribute if defined, else Y attribute. Or else literal, etc. Open to other names.

aaron-steinfeld:
This is ready for review now. Presumably there will be more changes to support persistence of this, but have some prototypes based on this that are getting ready to review.

aaron-steinfeld:
Going to merge this even though it may continue to change, so we can get the next parts of the POC up. May change later, so don't intend on anything outside of the POC using this yet.

PR #18, opened 2020-08-25
add attribute projections

aaron-steinfeld:
Will add tests before merging this, but trying to validate the concept first.


attribute-projections/src/main/java/org/hypertrace/core/attribute/service/projection/StringConcatenation.java
```diff
@@ -0,0 +1,35 @@
+package org.hypertrace.core.attribute.service.projection;
+
+import java.util.List;
+import org.hypertrace.core.attribute.service.v1.AttributeKind;
+
+public class StringConcatenation implements AttributeProjection {
+  public static StringConcatenation STRING_CONCATENATION = new StringConcatenation();
+
+  public static String concatenate(List<String> strings) {
```
tim-mwangi, line 9: Allow for a delimiter to be passed in?

tim-mwangi:
lgtm for the moment

attribute-projections/src/main/java/org/hypertrace/core/attribute/service/projection/AttributeProjection.java
```diff
@@ -0,0 +1,12 @@
+package org.hypertrace.core.attribute.service.projection;
+
+import java.util.List;
+import org.hypertrace.core.attribute.service.v1.AttributeKind;
+import org.hypertrace.core.attribute.service.v1.LiteralValue;
+
+public interface AttributeProjection {
+
+  AttributeKind getArgumentKindAtIndex(int index);
```
buchi-busireddy, line 9: Does this need to be exposed? Is it rather better to have a wrapper class for arguments so that the implementations can have metadata around the actual attribute also?

attribute-projections/src/main/java/org/hypertrace/core/attribute/service/projection/StringConcatenation.java
buchi-busireddy: Is there a delimiter?

aaron-steinfeld: Concatenate doesn't generally take a delimiter - join does. So three options:
1. Rename operator to join and add delimiter, which could be provided as (or potentially default to) empty
1. Two separate operators, join and concat with join requiring a delimiter
1. Leave it as is, because join is just syntax sugar on concat, and I can already define the args to this as any mix of literals/attributes/other projections.

What do you think?

attribute-projections/src/main/java/org/hypertrace/core/attribute/service/projection/AttributeProjection.java
aaron-steinfeld: I could potentially be convinced here, but reframing the question, it's still a matter of: Should the projection executor code (currently trace client) be responsible for coercing data to the right format, or should each projection get data in and be responsible for doing so internally. I tend to prefer the former because it feels like that coercing is more about data types than any specific implementation so would often be duplicated, but would gladly discuss.

buchi-busireddy: So, looking at operators (PROJECTION_OPERATOR_DEFAULT, PROJECTION_OPERATOR_CONCAT, etc), I thought the result type is implicit there, and hence the type might be internal to Projection was my understanding. It's true in first two cases where a Project is either an existing attribute or Literal because it knows the type. If ProjectionExpression also knows the type, everything will be uniform. wdyt?

attribute-projections/src/main/java/org/hypertrace/core/attribute/service/projection/StringConcatenation.java
tim-mwangi: 2. Clean and separate.

attribute-projections/src/main/java/org/hypertrace/core/attribute/service/projection/AttributeProjection.java
aaron-steinfeld: Not sure I follow - some operators will require strings to work with, but could take in a numeric attribute - for example if I wanted to hash a numeric and string together - will always produce a string value. There, the numeric input needs to be made into a string. Other operators, like DEFAULT don't care at all what the argument type is, so in those cases no conversion needs to be done/attempted. The return type depends on what comes in. In other words, you can default either a string or numeric field.

buchi-busireddy: My point is, project should know the incoming types because it has the arguments and their attribute metadata and it knows the output time so the conversion can be internal to it. Let me know if I'm missing something.

aaron-steinfeld: It certainly can be implemented internally, I just am missing what is gained by doing so. We have information on either side, so we'll need a back + forth communication either way. Currently, the executor 
1. Gets all the input values according to the definition - these can be a mixture of attributes that are pulled out of the span, literal values, and results of other projections.
2. Asks the operation what type it expects for each of those input values and converts as appropriate, erroring out if any can not be converted to the appropriate type. (this would go away in your suggestion, but be replaced by the same duplicated conversion inside of each operator)
3. Gets a result of unknown type and converts it into a literal value (likewise, this conversion could also be pushed down, but would be duplicated)

So the trade off that I see is: Is it better to take in make the functions `(literal[]) -> literal` (where literal is the literalvalue type currently in use that can hold a value of any type), and do the conversion inside each or `(object[]) -> object` (where object is the specific expected argument type/response type) and do all conversions in one spot outside. 

Ultimately, I would like to not expose the argument, but went with this approach for two reasons:
1. Just less repetitive. Even if sharing the coercion code itself, each projection gets more complex and those are the parts that I see changing/growing in the future.
2. In a future world (movie voice) where we are allowing defining attribute derivations on the fly, being able to lookup the metadata of the projection allows us to validate a definition and potentially build a UI around it. Admittedly, we'd need to build out more here potentially like number of args and return type which weren't needed right now.

aaron-steinfeld: I believe I addressed both of my concerns in the current iteration, and moved the conversions down lower.

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/Hash.java
```diff
@@ -0,0 +1,20 @@
...
+import com.github.f4b6a3.uuid.UuidCreator;
+import com.github.f4b6a3.uuid.creator.rfc4122.NameBasedSha1UuidCreator;
+import java.util.UUID;
+import javax.annotation.Nullable;
+
+public class Hash {
+
+  private static final NameBasedSha1UuidCreator HASHER =
+      UuidCreator.getNameBasedSha1Creator()
+          .withNamespace(UUID.fromString("5088c92d-5e9c-43f4-a35b-2589474d5642"));
```
tim-mwangi, line 14: is this one of those magic guids that should never change?

aaron-steinfeld: Yep - I planned on making this unique per tenant, but in the OLAP context we won't have that kind of information (unless provided explicitly). A UUIDv5 requires a namespace (actually, this lib seems to allow without but I don't know what the implications of that are). So this is just a randomly generated UUID - I didn't keep it in its own constant since by design it should never be referenced elsewhere.

Thoughts? Clearer as an explicit constant?

tim-mwangi: Actually as hardcoded it's better and all consumers can call this one method. Maybe some comments around it?

aaron-steinfeld: sure, will add a comment

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -54,6 +65,10 @@
   private List<AttributeSource> sources = Collections.emptyList();
   private Map<String, Map<String, String>> metadata = Collections.emptyMap();
 
+  @JsonSerialize(using = ProtobufMessageSerializer.class)
+  @JsonDeserialize(using = AttributeDefinitionDeserializer.class)
```
aaron-steinfeld, line 69: It works, but I'm definitely open to better ways fo accomplishing this if anyone has ideas. It seems much more error prone to manually trying to map the object and handle the compatibility here vs piggy backing off proto's existing support.

aaron-steinfeld:
Will add some more tests, but other than that I believe this is ready for review.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -340,6 +366,30 @@ public int hashCode() {
...
+    }
+  }
+
+  private static class AttributeDefinitionDeserializer extends JsonDeserializer<Message> {
+    private static final JsonFormat.Parser PARSER = JsonFormat.parser();
+
+    @Override
+    public Message deserialize(JsonParser parser, DeserializationContext context)
+        throws IOException {
+      AttributeDefinition.Builder builder = AttributeDefinition.newBuilder();
```
tim-mwangi, line 390: I think we can do something with Generics?

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
tim-mwangi: Never though of doing it this way. That said the serializer works for all types. The deserializer can be mad generic by using java generics I think?

aaron-steinfeld: Because the message contains the schema, we can be generic with the serializer. Coming from json (deser) however, we need to at a minimum know the target type (and use reflection) or better yet the builder class. I tried playing around with the deserialization context, but it doesn't appear to have anything helpful.

tim-mwangi: Yeah. Just tried it out. Doesn't work since the builder is constraining. :( Would have to spend some real time to see if it's doable another way.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
tim-mwangi: See comment above.

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -140,14 +140,14 @@ message Empty {
 
 message AttributeDefinition {
   oneof value {
-    string span_path = 1;
```
buchi-busireddy, line 143: Don't we need span path?

aaron-steinfeld: We will eventually, but only once we start using trace client. I'm parking that effort right now to focus on getting the straight projections working at the query layer.

attribute-projection-registry/src/main/java/org/hypertrace/core/attribute/service/projection/AttributeProjectionRegistry.java
```diff
@@ -0,0 +1,30 @@
...
+import org.hypertrace.core.attribute.service.projection.functions.Concatenate;
+import org.hypertrace.core.attribute.service.projection.functions.Hash;
+import org.hypertrace.core.attribute.service.v1.AttributeKind;
+import org.hypertrace.core.attribute.service.v1.ProjectionOperator;
+
+public class AttributeProjectionRegistry {
+
+  private static final Map<ProjectionOperator, AttributeProjection> PROJECTION_MAP =
+      Map.of(
+          PROJECTION_OPERATOR_CONCAT,
```
buchi-busireddy, line 17: I think we need to be able to chain the projections also right? For example, HASH(CONCAT(a, b)) so that we get the id automatically. That can come in later but it's better to keep them in mind.

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/DefaultValue.java
```diff
@@ -0,0 +1,13 @@
+package org.hypertrace.core.attribute.service.projection.functions;
+
+import static java.util.Objects.nonNull;
+
+import javax.annotation.Nullable;
+
+public class DefaultValue {
```
buchi-busireddy, line 7: Curious, where do we need this?

buchi-busireddy (approved):

attribute-projection-registry/src/main/java/org/hypertrace/core/attribute/service/projection/AttributeProjectionRegistry.java
aaron-steinfeld: Talked offline, but summary: yes we need and do support that, it's the responsibility of the layer above to resolve the arguments to literal values, which may include recursive projections.

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/DefaultValue.java
aaron-steinfeld: I removed the default function to try to focus on getting a POC with a single function first - left it in because we do want this functionality and I already had tested it, just didn't finish hooking it up. As an example of where this might be used, take the case where we're generating service name off the serviceName tag if provided, else something else. Our own elvis operator.

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/Hash.java
avinashkolluru: Is there any relation between this and the UUID generated in entity-service?

aaron-steinfeld: Good catch - I completely forgot that I shared the same namespace with that UUID generator. So right now, there's no relationship, but I shared the namespace because the plan is to eventually remove that generator and have things like entity IDs purely exist as projected attributes. The exact migration there is TBD.

buchi-busireddy: Does this mean we don't need to change entity-service id generation logic to be not fully dependent on identifying properties?

aaron-steinfeld: Confused by the double negative here, but the idea is that we're flipping the burden of ID generation. Right now, as I understand it, we upsert an entity by setting attribute values, on an entity object, sending it to entity service, and then reading back the ID from the response. Instead, the plan is to make this formation of an entity config driven rather than code driven - that is, rather than creating the entity explicitly by picking the identifying attributes, we'll have an attribute that is defined as the id attribute for an entity and can be read through the trace client. Entity formation/liveness can then be completely decoupled from enrichment, but use the same mechanism to get the ID. There's a number of TBDs here of course, but that's the general idea.

PR #23, opened 2020-09-11
feat: add get by id lookup to caching attr client

aaron-steinfeld:


caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/DefaultCachingAttributeClient.java
```diff
@@ -84,4 +98,25 @@ private NoSuchElementException buildErrorForMissingAttribute(String scope, Strin
...
+    this.scopeAndKeyLookup.put(
+        attributeMetadata.getId(),
+        new AttributeScopeAndKey(attributeMetadata.getScope().name(), attributeMetadata.getKey()));
+  }
+
+  private static final class AttributeScopeAndKey {
+    private final String scope;
+    private final String key;
+
+    private AttributeScopeAndKey(String scope, String key) {
```
tim-mwangi, line 117: Does this need an implementation of equals and hashCode?

aaron-steinfeld: Shouldn't, this isn't used as a key anywhere, only as a map value

tim-mwangi: ok.

tim-mwangi (approved):

PR #24, opened 2020-10-07
adding shutdown interface for gracefully closing channel

kotharironak:
The new interface will help in resolving the logging issue - https://github.com/hypertrace/hypertrace/issues/89

### DIFF
--- a/attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
+++ b/attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
@@ -20,15 +20,7 @@ import org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProvide
 /** Client for accessing the AttributeService This is a proto based implementation */
 public class AttributeServiceClient {
   private final AttributeServiceGrpc.AttributeServiceBlockingStub blockingStub;
-
-  public AttributeServiceClient(String host, int port) {
-    this(ManagedChannelBuilder.forAddress(host, port).usePlaintext().build());
-  }
-
-  public AttributeServiceClient(AttributeServiceClientConfig attributeServiceClientConfig) {
-    this(attributeServiceClientConfig.getHost(), attributeServiceClientConfig.getPort());
-  }
-
+  
   public AttributeServiceClient(Channel channel) {
     this.blockingStub =
         AttributeServiceGrpc.newBlockingStub(channel)
@@ -95,13 +87,5 @@ public class AttributeServiceClient {
   public void deleteSourceMetadata(
       Map<String, String> headers, AttributeSourceMetadataDeleteRequest req) {
     execute(headers, () -> blockingStub.deleteSourceMetadata(req));
-  }
-
-  public void shutDown() throws InterruptedException {
-    Channel channel = blockingStub.getChannel();
-    if(channel instanceof ManagedChannel) {
-      ManagedChannel managedChannel = (ManagedChannel) channel;
-      managedChannel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
-    }
   }
 }
--- a/attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
+++ b/attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
@@ -1,13 +1,9 @@
 package org.hypertrace.core.attribute.service.client;
 
 import io.grpc.Channel;
-import io.grpc.ManagedChannel;
-import io.grpc.ManagedChannelBuilder;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
-import org.hypertrace.core.attribute.service.client.config.AttributeServiceClientConfig;
 import org.hypertrace.core.attribute.service.v1.AttributeCreateRequest;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadataFilter;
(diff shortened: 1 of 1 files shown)


attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
```diff
@@ -94,4 +96,12 @@ public void deleteSourceMetadata(
       Map<String, String> headers, AttributeSourceMetadataDeleteRequest req) {
     execute(headers, () -> blockingStub.deleteSourceMetadata(req));
   }
+
+  public void shutDown() throws InterruptedException {
+    Channel channel = blockingStub.getChannel();
```
kotharironak, line 101: we have two public constructors including this one `AttributeServiceClient(Channel channel)`, so handling this way? Do you have any other suggestions?

tim-mwangi: Mmh. Never had to look into this. I guess you are the pioneer.

tim-mwangi (dismissed):

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
tim-mwangi: @aaron-steinfeld brought this to my attention https://github.com/hypertrace/hypertrace-core-graphql/blob/main/hypertrace-core-graphql-grpc-utils/src/main/java/org/hypertrace/core/graphql/utils/grpc/DefaultGrpcChannelRegistry.java

aaron-steinfeld: So we do this in gql - https://github.com/hypertrace/hypertrace-core-graphql/blob/main/hypertrace-core-graphql-grpc-utils/src/main/java/org/hypertrace/core/graphql/utils/grpc/DefaultGrpcChannelRegistry.java

I'd recommend removing the interfaces for creating a client without a channel - that way the server is responsible for calling and managing its channels rather than each client. That channel param can also likely be switched to a ManagedChannel.

kotharironak: The second part (That channel param can also likely be switched to a ManagedChannel) I got it, but in the first part, are you suggesting not to have `AttributeServiceClient(String host, int port)` and this one `AttributeServiceClient(AttributeServiceClientConfig attributeServiceClientConfig)`? 

And keeping `AttributeServiceClient(ManagedChannel channel)`, right?

kotharironak: I also need a similar `shutdown` interface for entity service related clients too -https://github.com/hypertrace/entity-service/tree/main/entity-service-client/src/main/java/org/hypertrace/entity

aaron-steinfeld: That would be my suggestion yes. Replacing the call is trivial enough (`AttributeServiceClient(ManagedChannelBuilder.forAddress...build())`, but it (intentionally) puts the burden on the caller to manage their channels, rather than the client. The client then can be agnostic to things like channel lifecycle and reuse which are service-specific.

aaron-steinfeld: A little about grpc channel best practices here: https://groups.google.com/g/grpc-io/c/jc-cZak29V4

aaron-steinfeld: Specifically, the discussion about cannel reuse:

> Reading another thread made me want to tweak my statement:
> Only in very heavy throughput (when we start talking about trying to fully saturate a server NIC) should you even begin to consider do something else.
> 
> Even in that case you probably shouldn't use multiple Channels. Instead, you would want a load balancer that might make multiple connections to the backend. Use different channels when you are contacting different backends. That is the only form of "channel pool" you might want: to look up a single shared Channel per backend. .

So I'm not saying we need to do that as part of this change at all (that's a service rather than client change anyway), but by taking a channel as the client builder argument, it encourages consuming services to better manage their channels rather than treating them as an implementation detail of the client.

aaron-steinfeld: > I also need a similar `shutdown` interface for entity service related clients too -https://github.com/hypertrace/entity-service/tree/main/entity-service-client/src/main/java/org/hypertrace/entity

All clients do need this change, but the shutdown interface isn't needed if we pass in the channel - it would be shutdown from outside.

kotharironak: Now, I got it (was confused with the word `server`). I will make changes, and update the PR.

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
```diff
@@ -1,10 +1,12 @@
 package org.hypertrace.core.attribute.service.client;
 
 import io.grpc.Channel;
+import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
```
aaron-steinfeld, line 9: nit: extra imports

aaron-steinfeld (approved):

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceClient.java
kotharironak: done.

aaron-steinfeld (approved):

tim-mwangi (approved):

PR #26, opened 2020-10-16
feat: add get attributes in scope to client api

aaron-steinfeld:


attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -99,6 +99,7 @@ message AttributeMetadata {
   string id = 14;
   bool groupable = 15;
   AttributeDefinition definition = 16;
+  string scope_string = 17;
```
aaron-steinfeld, line 102: Starting to replace the enum scope with a string version to remove requirement of compile-time knowledge of scopes.

buchi-busireddy (approved):

PR #27, opened 2020-10-19
feat: make scope a string, deprecate enum

aaron-steinfeld:


attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModelTest.java
```diff
@@ -256,4 +253,108 @@ public void testAttributeDefinitionBackwardsCompatibility() throws IOException {
...
+            + "}";
+    expected =
+        template.toBuilder()
+                .setScope(AttributeScope.EVENT)
+                .setScopeString("EVENT")
+                .setId("EVENT.key")
+                .build();
+    Assertions.assertEquals(expected, AttributeMetadataModel.fromJson(inputJson).toDTO());
+
+    // Given JSON with unknown scope string, scope string only should be set
```
tim-mwangi, line 343: what about both scope and scope_string? is that an important path?

tim-mwangi (dismissed):

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -366,7 +357,8 @@ public boolean equals(Object o) {
         && Objects.equals(supportedAggregations, that.supportedAggregations)
         && Objects.equals(sources, that.sources)
         && Objects.equals(metadata, that.metadata)
-        && Objects.equals(definition, that.definition);
+        && Objects.equals(definition, that.definition)
```
tim-mwangi, line 360: There's half codecov for this. It's yellow instead of green

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModelTest.java
aaron-steinfeld: I'd guess the one defined later would win. It definitely wouldn't break. 
Regardless, it's not an important path - the json schema is changed with this change - scope can be read from the database but no longer written (ln 155 of AttributeMetadataMode). So existing entries would read a scope and set both values of an attribute (case on line 325 here), anything being written (or updated) after this change would write only scope_string.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
aaron-steinfeld: I'd guess that's because this is a ton of branches, and we're not writing tests for each permutation of model equality so it never gets to this bottom branch. Do you think it's worth writing tests for them all considering this method is generated? Also from a quick peak around, I don't believe equal/hashcode are used on this model anyway.

tim-mwangi: ok. I always shoot for 100% but i guess it's not crucial in this case.

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModelTest.java
tim-mwangi: ok

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -328,7 +332,11 @@ private void sendResult(
 
   private Query getQueryForFilter(
       String tenantId, AttributeMetadataFilter attributeMetadataFilter) {
-    List<AttributeScope> scopeFilterRequest = attributeMetadataFilter.getScopeList();
+    List<String> scopeFilterRequest =
```
buchi-busireddy, line 335: nit: `s/scopeFilterRequest/scopeFilterList/`?

buchi-busireddy (dismissed):

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
aaron-steinfeld: I noticed that, commented to myself that the existing naming was bad, and left it as is. Let me fix.

buchi-busireddy (approved):

PR #30, opened 2020-11-11
updates release workflow

JBAhire:
addresses https://github.com/hypertrace/hypertrace/issues/19 & https://github.com/hypertrace/hypertrace/issues/22


.circleci/config.yml
```diff
@@ -148,22 +152,33 @@ jobs:
...
-            CHART_VERSION=$(git describe --abbrev=0)
-            CHART_NAME=$(awk '/^name:/ {print $2}' ./helm/Chart.yaml)
-            export GOOGLE_APPLICATION_CREDENTIALS=${HOME}/helm-gcs-key.json
-            echo ${HELM_GCS_CREDENTIALS} > ${GOOGLE_APPLICATION_CREDENTIALS}
-            helm dependency update ./helm/
-            helm repo add helm-gcs ${HELM_GCS_REPOSITORY}
-            helm package --version ${CHART_VERSION} --app-version ${CHART_VERSION} ./helm/
-            helm gcs push ${CHART_NAME}-${CHART_VERSION}.tgz helm-gcs --public --retry
-
+            .circleci/helm.sh package
```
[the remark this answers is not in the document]
JBAhire: okay!

.circleci/config.yml
```diff
@@ -43,7 +44,7 @@ commands:
       - checkout
       - run:
           name: Generate cache key
-          command: find . -type f -name "*.gradle*" -exec shasum {} + | sort > /tmp/checksum.txt && cat /tmp/checksum.txt
+          command: find . -type f -name "*.gradle-wrapper*" -exec shasum {} + | sort > /tmp/checksum.txt && cat /tmp/checksum.txt
```
aaron-steinfeld, line 47: This change will cause cache misses, no files will match this pattern and the cache will never be updated. If it's being propagated from another place, that should be fixed. The intent is to key off any changes in the gradle code or the gradle (wrapper) version.

 The most accurate version of this command is probably
```bash
find . -type f -name "*.gradle*" -o -name "gradle-wrapper*" | xargs shasum | sort > /tmp/checksum.txt && cat /tmp/checksum.txt
```
Which produces the below on a test repo
```
33b0acb4572934fc1dd7d5880c65036724974e06  ./gradle/wrapper/gradle-wrapper.jar
6e2b9651ce90b9a8de1304134e4c0ef81ecfae80  ./service-impl/build.gradle.kts
723f2062a71251313edc9e6c3cadce31da8f89ed  ./settings.gradle.kts
96abe7707ce331aa7b3cf1961fe8ebe9e812c679  ./gradle/wrapper/gradle-wrapper.properties
a9e12dd6e753e9ba5a360ce1cf79d3afcaeacd94  ./service/build.gradle.kts
aed929c4fb91822a2ed57268c91dd60cafbf4c64  ./build.gradle.kts
bf3a8361aea58099052e4b2a4db0968da70568d0  ./service-api/build.gradle.kts
e47aaf32c16b04267fd91d0c183f9526eb62ab0c  ./semantic-build-versioning.gradle
```

.circleci/config.yml
```diff
@@ -181,26 +202,35 @@ workflows:
...
             - hypertrace-vulnerability-scanning
             - dockerhub-read
-      - publish:
+      - merge-publish:
           context:
             - hypertrace-publishing
             - dockerhub-read
           requires:
             - build
-            - integration-test
```
aaron-steinfeld, line 190: This should be a publish prerequisite, but appears to be removed

.circleci/config.yml
```diff
@@ -148,12 +152,23 @@ jobs:
...
-  package-charts:
+  release-charts:
     executor: helm
+    working_directory: ~/repo
     steps:
       - checkout
+      - run:
+          name: Add release tag
+          command: |
+            git config --global user.email "${CIRCLE_USERNAME}@hypertrace.org"
```
aaron-steinfeld, line 163: We generally tag before publishing - this appears to tag in between publish jars/images and helm charts.

.circleci/config.yml
```diff
@@ -163,7 +178,13 @@ jobs:
             helm repo add helm-gcs ${HELM_GCS_REPOSITORY}
             helm package --version ${CHART_VERSION} --app-version ${CHART_VERSION} ./helm/
             helm gcs push ${CHART_NAME}-${CHART_VERSION}.tgz helm-gcs --public --retry
-
+      - add_ssh_keys:
+          fingerprints:
+            # This ssh key gives write permission needed for the following step.
+            - '76:7a:c8:91:c0:cd:eb:71:96:3c:44:c6:cf:b0:26:83'
```
aaron-steinfeld, line 184: This appears to be a different fingerprint - intentional?

buchi-busireddy:
@JBAhire are you planning to address @aaron-steinfeld's comments? I think a similar PR being merged in gateway is causing failures on publish.

JBAhire:
> @JBAhire are you planning to address @aaron-steinfeld's comments? I think a similar PR being merged in gateway is causing failures on publish.

I somehow missed this @buchi-busireddy. 

@aaron-steinfeld , Thanks for the review. will address your comments. :)

.circleci/config.yml
JBAhire: got it.

JBAhire: will check if this is the case somewhere else as well.

.circleci/config.yml
JBAhire: my bad. corrected it.

.circleci/config.yml
JBAhire: addressed.

.circleci/config.yml
JBAhire: Yes, you are right. 

But at other places, we first tag, publish jars/images, and push the tags, and then upload helm charts.

So, do you suggest first tag, and than upload everything and publish tags after helm charts pushed.

For this PR, I had followed the workflow from hypertrace-ingester (https://github.com/hypertrace/hypertrace-ingester/blob/main/.circleci/config.yml) and pinot (https://github.com/hypertrace/pinot/blob/main/.circleci/config.yml) etc. which follow similar approach as this one.

JBAhire:
@aaron-steinfeld , addressed your comments. Please review and let me know if any more changes are required. If this looks good, I will follow similar pattern across all repos as this will help us in our next important goal i.e. putting in proper release process.

.circleci/config.yml
```diff
@@ -43,7 +44,7 @@ commands:
       - checkout
       - run:
           name: Generate cache key
-          command: find . -type f -name "*.gradle*" -exec shasum {} + | sort > /tmp/checksum.txt && cat /tmp/checksum.txt
+          command: find . -type f -name "*.gradle*" -o -name "gradle-wrapper*" | xargs shasum | sort > /tmp/checksum.txt && cat /tmp/checksum.txt
```
jcchavezs, line 47: This looks quite sophisticated, I am trying to read through the code what is this doing but not much success. Do you mind adding a comment saying: `# find x | shasum y | sort by x`

.circleci/config.yml
aaron-steinfeld: This is fine as is, thanks for jogging my memory - for context, when we were first implementing the release workflow we ran into issue with the multi-stage publish process. Because each job in the pipeline is filtered to a certain tag pattern (so it doesn't run on a standard merge), that tag needs to remain through each stage, and we swap in the new tag at the end.  There were a few other alternatives and their problems discussed here: https://github.com/hypertrace/pinot/pull/19#discussion_r481179127

Changing the local tags before publishing the chart is just a bit of a shortcut so we don't have to parse the tag to get the chart version, but they're not actually pushed until the last step.

.circleci/config.yml
```diff
@@ -148,12 +152,23 @@ jobs:
       - setup_build_environment
       - snyk/scan:
           additional-arguments: --all-sub-projects --policy-path=.snyk
-  package-charts:
+  release-charts:
     executor: helm
+    working_directory: ~/repo
```
aaron-steinfeld, line 157: working directory not needed, we generally use the default

.circleci/config.yml
aaron-steinfeld:
```suggestion
          # Find all gradle related files, hash each and order them consistently, then save to file for later use
          command: find . -type f -name "*.gradle*" -o -name "gradle-wrapper*" | xargs shasum | sort > /tmp/checksum.txt && cat /tmp/checksum.txt
```

aaron-steinfeld (commented):
One missing piece of this PR is updating the docker plugin versions - see the changelog https://github.com/hypertrace/hypertrace-gradle-docker-plugins/blob/main/hypertrace-gradle-docker-plugin/CHANGELOG.md (really need to reverse that order). The important part for the release process is the change in tagging that came in the 0.6.x release, but may as well bring it up to latest. I saw this issue in gateway service, where several snapshot versions were published to github which required manual clean up.

.circleci/config.yml
JBAhire: removing this

aaron-steinfeld (approved):

tim-mwangi (approved):

PR #31, opened 2020-11-12
feat: add conditional and strings equal projection

tim-mwangi:
## Description
Add conditional and string equals projections. Also need a TernaryAttributeProjection for functions that take in 3 arguments.

### Testing
Unit tests for all the methods added.

### Checklist:
- [✅ ] My changes generate no new warnings
- [✅  ] I have added tests that prove my fix is effective or that my feature works


tim-mwangi:
@aaron-steinfeld copied and pasted your idea. Let me know if this is in the right direction before I make the changes in Pinot as well.

attribute-projection-registry/src/main/java/org/hypertrace/core/attribute/service/projection/TriFunction.java
```diff
@@ -0,0 +1,13 @@
+package org.hypertrace.core.attribute.service.projection;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+public interface TriFunction<T, U, V, R> {
+  R apply(T t, U u, V v);
+
+  default <W> TriFunction<T, U, V, W> andThen(Function<? super R, ? extends W> after) {
```
aaron-steinfeld, line 9: just for fun? or are we using andThen anywhere?

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/Equals.java
```diff
@@ -0,0 +1,17 @@
...
+import javax.annotation.Nullable;
+
+public class Equals {
+  private static final String DEFAULT_STRING = "";
+
+  public static boolean stringEquals(@Nullable String first, @Nullable String second) {
+    if (isNull(first) && isNull(second)) {
+      return true;
+    }
+    return requireNonNullElse(first, DEFAULT_STRING).equals(requireNonNullElse(second, DEFAULT_STRING));
```
aaron-steinfeld, line 15: so this mans that `null === ""` - not sure that should hold. Could we do
```java
if (isNull(first) && isNull(second)) {
      return true;
 }
if (isNull(first) || isNull(second)) {
      return false;
 }
return first.equals(second);
```

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/Conditional.java
```diff
@@ -0,0 +1,13 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class Conditional {
+  @Nullable
+  public static String getValue(@Nonnull boolean condition,
+                                @Nullable String first,
+                                @Nullable String second) {
+    return condition ? first : second;
```
aaron-steinfeld, line 11: I think we should probably always support nullable args. If condition is null, we can return null - but we never want to NPE.

attribute-projection-registry/src/main/java/org/hypertrace/core/attribute/service/projection/TriFunction.java
tim-mwangi: Oh I copied and modified Java's BiFunction code haha

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/Equals.java
tim-mwangi: So no defaulting to empty string if null? Got it.

attribute-projection-functions/src/main/java/org/hypertrace/core/attribute/service/projection/functions/Conditional.java
tim-mwangi: ok. got it. bad things happen to good people.

tim-mwangi:
@aaron-steinfeld what do you think of the class names?

aaron-steinfeld:
They seem fine to me - I don't love string equals, but without overload support (which pinot doesn't have for UDFs, so we didn't build for our in-memory projections) I think we're stuck on our equals naming/implementations.

attribute-projection-registry/src/main/java/org/hypertrace/core/attribute/service/projection/TriFunction.java
tim-mwangi: https://media.giphy.com/media/3OfNDMhTCi7zW/giphy.gif

aaron-steinfeld (approved):

buchi-busireddy (approved):

PR #32, opened 2020-11-18
feat: Upgrading docstore version to upgrade mongo java driver.

buchi-busireddy:
## Description
Upgrading the docstore version to latest, which changes the mongo java driver used from 3.12 to 4.1.1. This is because the previous java driver is very old, marked legacy driver by MongoDB and there are a lot of bug fixes in the newer versions.

Most importantly, we are running into some issues with bulk upsert in production and we're trying to see if we see those same issues with the latest Mongo java driver too.

### Testing
Ran Hypertrace locally with the locally built image and things are working fine.

### Checklist:

### Documentation
No functionality is changed, hence no docs needed.

### DIFF
--- a/attribute-service-impl/build.gradle.kts
+++ b/attribute-service-impl/build.gradle.kts
@@ -8,7 +8,7 @@ dependencies {
   api(project(":attribute-service-api"))
   implementation(project(":attribute-service-tenant-api"))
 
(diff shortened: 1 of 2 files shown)


buchi-busireddy:
Very weird, I didn't add any tests in this PR but the coverage has gone up :)

tim-mwangi (approved):

PR #39, opened 2020-11-24
adding qoute for supporting colon in url

kotharironak:
## Description
fix the qoute issue to supports url with a colon for document store configuration 
e.g "mongodb://mongo:27017"

HOCON treats `:` as a value separator (https://github.com/lightbend/config/blob/master/HOCON.md#key-value-separator), fixing the parsing error by quoting each value string.

### Testing
Tried locally with - `helm template ./helm/` by adding above `url` values in values.yaml file.
e.g output:
```
application.conf: |-
    document.store {
      dataStoreType = mongo
      mongo {
        host = "mongo"
        url = "mongodb://mongo:27017"
      }
    }
```
### Checklist:

### Documentation
Not required

### DIFF
--- a/helm/templates/configmap.yaml
+++ b/helm/templates/configmap.yaml
@@ -12,7 +12,7 @@ data:
       {{ $dst }} {
         {{- range $key, $value := (index .Values "configMap" (printf "%s" $dst)) }}
         {{- if $value }}
-        {{ $key }} = {{ $value }}
+        {{ $key }} = {{ $value | quote }}
         {{- end }}
         {{- end }}
       }


ravisingal (approved):

helm/templates/configmap.yaml
```diff
@@ -12,7 +12,7 @@ data:
       {{ $dst }} {
         {{- range $key, $value := (index .Values "configMap" (printf "%s" $dst)) }}
         {{- if $value }}
-        {{ $key }} = {{ $value }}
+        {{ $key }} = {{ $value | quote }}
```
aaron-steinfeld, line 15: This will quote ports too right? Is that a problem?

kotharironak: I also had the same concern (https://github.com/lightbend/config/blob/master/HOCON.md#automatic-type-conversions). So, we tested by making changes locally.

avinashkolluru (approved):

helm/templates/configmap.yaml
aaron-steinfeld: Gotcha - saw port missing in the PR description, and thought it might have been omitted in the verification :)

aaron-steinfeld (approved):

PR #45, opened 2020-12-18
Fixed merge-publish and disabled circleci publish workflow

JBAhire:
## Description
- Addresses Ronak's comments here: https://github.com/hypertrace/attribute-service/pull/43#discussion_r545650645 
- Merges jobs into 2 workflow files. 
- Disables CircleCI `merge-publish` and `release-publish` workflow.
- Merge-publish didn't run because workflow runs on push to main and not on PR closed so it doesn't have merge context. For merge condition to run the action should be `closed` and the merged key be true (as per: https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request). So, changed condition and also triggers for PR. 

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:

### Documentation
- https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request).
- https://github.community/t/trigger-workflow-only-on-pull-request-merge/17359/6
- https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request


.github/workflows/pr-build.yml
```diff
@@ -61,8 +63,42 @@ jobs:
           flags: integration
 
       - name: push docker image
-        if: github.event.pull_request.merged
+         # https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request
+        if: github.event_name == 'pull_request' && github.event.action == 'closed' && github.event.pull_request.merged == true
         run: ./gradlew dockerPushImages
 
+  validate-helm-charts:
```
aaron-steinfeld, line 70: @kotharironak - I disagree with this consolidation.

Keeping the jobs separate has a few different benefits
- They run in parallel, giving a faster time to green build
- The reported results are more granular, making it easier for the developer to figure out what went wrong
- The jobs are kept independent and don't unintentionally use state from a previous step
- Developers get more feedback because a serial run would stop at the first failure, while a parallel run gives the results for each

The trade offs that I see are only:
- Takes more action minutes (not an issue at the moment, this is a free project)
- More maintenance (I actually think keeping workflows simple is actually easier to maintain even if more files/LOC)

.github/workflows/pr-build.yml
```diff
@@ -61,8 +63,42 @@ jobs:
           flags: integration
 
       - name: push docker image
-        if: github.event.pull_request.merged
+         # https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#pull_request
+        if: github.event_name == 'pull_request' && github.event.action == 'closed' && github.event.pull_request.merged == true
```
aaron-steinfeld, line 67: Discussed offline with @JBAhire , rather than getting fancy with the conditions and triggers which seems more fragile, we can just split the workflows up to not share triggers - the extra job is justified IMO (explained more below)

.github/workflows/pr-build.yml
kotharironak: As per github actions docs,
- jobs run in parallel if the workflow has multiple jobs- https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#jobs
- in Github hosted runner each job run in fresh virtual envs - https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#runners
So, I think combing multiple jobs that trigger on the same condition will be easy to manage directly from the action tab. With the above changes - https://github.com/hypertrace/attribute-service/actions/runs/430169794 (Total time ~2m)

Below was the previous case where we had three workflows on the same triggers and three things to look at (Total time ~3min)
pr-build.yml - https://github.com/hypertrace/attribute-service/actions/runs/430001233
synk.yml - https://github.com/hypertrace/attribute-service/actions/runs/430001235
validate-charts.yml - https://github.com/hypertrace/attribute-service/actions/runs/430001234

.github/workflows/pr-build.yml
```diff
@@ -61,8 +63,42 @@ jobs:
...
+    container: 
+      image: hypertrace/helm-gcs-packager:0.3.0
+      credentials:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+     # Set fetch-depth: 0 to fetch commit history and tags for use in version calculation
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2.3.4
+        with:
```
kotharironak, line 81: We will need the below condition to take for the forked pull request otherwise it will checkout only the `main` branch code
```
with:
        ref: ${{github.event.pull_request.head.ref}}
        repository: ${{github.event.pull_request.head.repo.full_name}}
```

.github/workflows/pr-build.yml
aaron-steinfeld: Sorry I misunderstood the ask - I thought we were combining jobs, not workflows. Combining the workflows based on the triggers makes sense, so disregard. Really what I was looking for is that they show up independently on the PR itself, which they appear to be in this change. I think optimizing those screens - that is 
```
https://github.com/hypertrace/attribute-service/pull/45
https://github.com/hypertrace/attribute-service/pull/45/checks
```
is the much more common use case vs looking at the repo-wide action runs (but sounds like we've got both!)

aaron-steinfeld: Given this though, does it make sense to rename the workflows based on those triggers rather than their content?

kotharironak: Yes, that is what I was suggesting @JBAhire  as well. I was also suggesting that instead of running both the targets on each pull request (`pull_request_target` and `pull_request`) we can have three workflows.

1. pr-build-forked.yml (runs only for forked branches) - in `pull_request_target` context
2. pr-build.yml (runs only for member's pull request) - in `pull_request` context
3. publish.yml (only for member pull request) - which is anyway the case as it triggers on `release`

Most of the steps will be the same (and may look the same copy) between the `pr-build-forked` and `pr-build` workflow, but, we will able control what goes in forked PRs workflow. And, makes it easy to separate issues (like secrets or anything).

aaron-steinfeld: Doesn't `pull_request_target` run on branch + fork PRs? (for example, it ran on this PR which is a branch). I was thinking of getting rid of the `pull_request` trigger entirely, and instead having
- `pull_request_target` - everything we expect on a PR, which should be the same for members or contributors
- `push` - everything we expect to run on merge (basically, the "merge-publish" from before - only triggered by members who merge)
- `release` - everything for cutting a release (only members who cut a release would be able to trigger)

.github/workflows/pr-build.yml
JBAhire: yup. forgot about this earlier

.github/workflows/pr-build.yml
JBAhire: so there are 2 problems with using `pull_request_target` for this workflow.
- codecov action error which you can see here as well. (https://github.com/codecov/codecov-action/issues/155)
- checkout like this: https://github.com/hypertrace/attribute-service/pull/45#discussion_r545827000

JBAhire: ideally only `pull_request_target` would be enough for most of the situations and that's what many projects follow, trigger on `push` and `pull_request_target`.

aaron-steinfeld: - checkout: This is just a change in the checkout command right? Does it have other implications or is that resolved once we make that change?

- codecov: If `pull_request_target` doesn't work for upload, would it make sense to create a fourth workflow
`pull_request` with one job, `test` that runs the tests and uploads coverage. That shouldn't need secrets, right?

aaron-steinfeld: Basically, I'm interested if we can avoid duplicating between the `pull_request_target` and `pull_request` workflows, and just sort the jobs based on what they need in terms of available info.

aaron-steinfeld:
Another issue I forgot til now - the tag generation for docker non-release images is based off the circle ci branch - we should update that: 
Ref: https://github.com/hypertrace/hypertrace-gradle-docker-plugins/blob/c4cbaa615a03724c17c9eed0decf7311ac8128dc/hypertrace-gradle-docker-plugin/src/main/java/org/hypertrace/gradle/docker/DockerPlugin.java#L111-L119

kotharironak:
> https://github.com/hypertrace/hypertrace-gradle-docker-plugins/blob/c4cbaa615a03724c17c9eed0decf7311ac8128dc/hypertrace-gradle-docker-plugin/src/main/java/org/hypertrace/gradle/docker/DockerPlugin.java#L111-L119

As we will have to publish docker artifacts on `main`, we can do that for the `on push` event trigger of the `main` branch.  Based on this doc -
https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context, we can derive it from `GITHUB_REF` for the above `on push` event.

we will have to expose a branch `GITHUB_BRANCH` to the `merge-publish` step, so it can be read in the gradle plugin. And we can do this by adding one more step like below before `merge-publish`
```
if: github.event_name == 'push' 
	 run: | 
	 GHA_BRANCH="${GITHUB_REF#refs/heads/}"
	 echo "GHA_BRANCH=${BRANCH}" >> $GITHUB_ENV
```

.github/workflows/pr-build.yml
kotharironak: The reason I was suggesting above (forked vs non-forked flow)
a. avoiding running both (pull_request, pull_request_target) on the same PR (but your proposal can also handle it)
b. educating folks (members) on `pull_request_target` and `pull_request` for security, and if some repo's are having fewer contributions, we are still running most of the PRs in `pull_request_target`
c. If we don't want to run certain steps in `forked` PR like Snyk (this was under discussion), it can help.
d. resource utilization (`test` workflow anyway depends on `build`) and better workflow view (https://github.com/hypertrace/attribute-service/actions/runs/430169794)

But, I am fine with your suggestion too. That will also help in codcov issue for now which we might have to omit in (pr-build-forked, pr-build) proposal.

Just to summerize:
pr-build - on `pull_reqeust_target` + `push`
pr-test - on `pull_reqeuest` + `push`
merge_publish - on push
publish - on release tag created

JBAhire: will create this 4 workflows.

.github/workflows/merge-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+        
+      - name: branch name tag
+        run: | 
+          BRANCH="${GITHUB_REF#refs/heads/}"
```
aaron-steinfeld, line 41: Since this is a standard ref format, can we make the plugin handle parsing it rather than needing to do it in each job (also avoids creating a custom env variable that our plugin has knowledge of).

In other words, have the plugin look for either CIRCLE_BRANCH or GITHUB_REF, then check if it's in fully qualified or shorthand format to get the branch name.

.github/workflows/publish.yml
```diff
@@ -51,20 +51,14 @@ jobs:
...
-          password: ${{ secrets.DOCKERHUB_TOKEN }}       
    
       - name: package and release charts
         env:
           HELM_GCS_CREDENTIALS: ${{ secrets.HELM_GCS_CREDENTIALS }}
           HELM_GCS_REPOSITORY: ${{ secrets.HELM_GCS_REPOSITORY }}
             # GITHUB_REF will be refs/tags/docker-MAJOR.MINOR.PATCH
         run: |
-            CHART_VERSION=$(git describe --abbrev=0)
+            CHART_VERSION=$(echo ${GITHUB_REF} | cut -d/ -f 3)
```
aaron-steinfeld, line 61: Does the tag ref not come down here? In other words, the old command seemed simpler unless there's a reason to change

.github/workflows/pr-build.yml
```diff
@@ -38,31 +40,42 @@ jobs:
...
-          fail_ci_if_error: true
-          verbose: true
-          flags: integration
-
-      - name: push docker image
-        if: github.event.pull_request.merged
-        run: ./gradlew dockerPushImages
-
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
```
aaron-steinfeld, line 73: So I wrote the below comment, and then realized that the jobs are already sorted this way. So I think we should make the change in naming to reflect the actual distinction in the workflows, the triggers (currently the build job has several validation/test tasks):


> Should we make the `pull_request` workflow our default location for jobs, and limit `pull_request_target` to the small set of jobs that need access to secrets? The benefits I was thinking were that
> - The risk of accidental secret exposure decreases (because we're running less jobs in forked contexts)
> - The extra dance around checkout is needed on fewer jobs
> The drawbacks being-
> - The workflow would be sorted/named by trigger than than function (but the results are generally reported by job, not workflow)
>

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.gradle
+          key: gradle-packages-${{ runner.os }}-${{ github.job }}-${{ hashFiles('**/checksum.txt') }}
+          restore-keys: |
+            gradle-packages-${{ runner.os }}-${{ github.job }}
+            gradle-packages-${{ runner.os }}
+
+      - name: Unit test
+        run: ./gradlew jacocoTestReport
```
aaron-steinfeld, line 33: nit: is there a way of displaying test results in GHA - CCI has a way of uploading test results and I'd guess there's some equivalent for GHA. At a minimum, uploading the junit html report as a test artifact, if nothing first class for tests.

JBAhire: yup. we can upload artifacts.

JBAhire:
> > https://github.com/hypertrace/hypertrace-gradle-docker-plugins/blob/c4cbaa615a03724c17c9eed0decf7311ac8128dc/hypertrace-gradle-docker-plugin/src/main/java/org/hypertrace/gradle/docker/DockerPlugin.java#L111-L119
> 
> As we will have to publish docker artifacts on `main`, we can do that for the `on push` event trigger of the `main` branch. Based on this doc -
> https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context, we can derive it from `GITHUB_REF` for the above `on push` event.
> 
> we will have to expose a branch `GITHUB_BRANCH` to the `merge-publish` step, so it can be read in the gradle plugin. And we can do this by adding one more step like below before `merge-publish`
> 
> ```
> if: github.event_name == 'push' 
> 	 run: | 
> 	 GHA_BRANCH="${GITHUB_REF#refs/heads/}"
> 	 echo "GHA_BRANCH=${BRANCH}" >> $GITHUB_ENV
> ```

addressed by https://github.com/hypertrace/hypertrace-gradle-docker-plugins/pull/21

.github/workflows/publish.yml
JBAhire: I am not sure about this one. Let's use the earlier one and if any issue is there we can switch. I found most of the workflows using GITHUB_REF way

.github/workflows/pr-build.yml
JBAhire: will change the name

.github/workflows/pr-test.yml
```diff
@@ -39,9 +42,19 @@ jobs:
...
+      - name: Archive unit test report
+        uses: actions/upload-artifact@v1
+        with:
+          name: unit-test
+          path: /unit-test-reports
+        if: ${{ always() }}
 
       - name: Integration test
-        run: ./gradlew jacocoIntegrationTestReport
+        run: ./gradlew jacocoIntegrationTestReport copyAllReports --output-dir=/integration-test-reports
```
aaron-steinfeld, line 54: copy all reports should be called in next step only

.github/workflows/pr-test.yml
```diff
@@ -39,9 +42,19 @@ jobs:
...
+
+      - name: Copy integration test reports
+        run: ./gradlew copyAllReports --output-dir=/tmp/integration-test-reports
```
aaron-steinfeld, line 57: this output dir doesn't match the upload directory on line 71

JBAhire: fixed this and everything is working now: https://github.com/Fininity-ai/Attributes-test/runs/1586845665

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          name: unit-test
+          path: /tmp/unit-test-reports
+        if: ${{ always() }}
+
+      - name: Integration test
+        run: ./gradlew jacocoIntegrationTestReport 
+
+      - name: Copy integration test reports
+        run: ./gradlew copyAllReports --output-dir=/tmp/integration-test-reports
```
aaron-steinfeld, line 57: The artifacts produced by this are the same as the artifacts reported by the previous step I believe since we're not cleaning out. We can either clean between the two test phases, or a single upload for both runs is probably fine too.

As a next step (can be a separate PR) it would make things much nicer to extract the results so they're visible in the GH UI (looks like it doesn't let viewing the artifacts, only downloading them). Something like this action: 
https://github.com/marketplace/actions/publish-unit-test-results

aaron-steinfeld (approved):

buchi-busireddy (approved):

.github/workflows/pr-test.yml
JBAhire: Noted. Makes sense to me.

.github/workflows/pr-test.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          name: codecov-line-smoothing
+          fail_ci_if_error: true
+          verbose: true
+          flags: unit
+      
+      - name: Archive unit test report
```
kotharironak, line 46: Why do we have to archive?

JBAhire: https://github.com/hypertrace/attribute-service/pull/45#discussion_r546703204

.github/workflows/merge-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+            gradle-packages-${{ runner.os }}-${{ github.job }}
+            gradle-packages-${{ runner.os }}
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+        
+      - name: branch name tag
```
kotharironak, line 39: We don't need this now as you moved the extraction of BRANCH_NAME in the Gradle plugin - https://github.com/hypertrace/hypertrace-gradle-docker-plugins/pull/21/files?

.github/workflows/merge-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          fetch-depth: 0
+      
+      - name: create checksum file
+        run: find . -type f -name "*.gradle*" -o -name "gradle-wrapper*" | xargs shasum | sort > checksum.txt && cat checksum.txt
+
+      - name: Cache packages
+        id: cache-packages
+        uses: actions/cache@v2
```
kotharironak, line 25: Let's use the same version `v2.3.4` everywhere.

kotharironak (approved):
I have a few minor comments, can you take care of it in follow up PR if you think they are valid.  
LGTM

JBAhire:
@kotharironak , will do that.

PR #48, opened 2020-12-22
avoids exposing env variables in pr-build

JBAhire:
## Description
- Avoids exposing secrets in `pr-build` which runs in `pull_request_target` context. 
- adds publish-images dependency for helm chart publish job. 
- Adds codeowners for github actions

### Checklist:

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -5,7 +5,6 @@ on:
       - main
   pull_request:
 
-
 jobs:
   test:
     runs-on: ubuntu-20.04
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,3 +2,6 @@
 
 # global
 * @buchi-busireddy @avinashkolluru @tim-mwangi
+
+# GH action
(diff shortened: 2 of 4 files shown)


aaron-steinfeld (approved):

aaron-steinfeld (dismissed):

kotharironak (approved):

.github/workflows/pr-test.yml
```diff
@@ -5,7 +5,6 @@ on:
       - main
   pull_request:
 
-
```
kotharironak, line 8: Looks like no change.

kotharironak (approved):

aaron-steinfeld (approved):

.github/workflows/pr-test.yml
JBAhire: ahh. I was planning to add branch restriction here as well. maybe that line got removed. no worries

aaron-steinfeld: hmm - it might be good to have symmetry with the `pull_request_target` workflow, but on the other hand, raising a PR against a non-main branch is a fine practice, so no harm done by leaving it unfiltered.

PR #54, opened 2020-12-23
switches to GITHUB_REF for tagging helm charts

JBAhire:
## Description
We are using `git describe --abbrev=0` for tagging charts but it's somehow is still on `0.8.7` and we are not able to release helm charts because of that. 

With this PR, we are using GITHUB_REF to get tag for helm charts as well which is standard practice.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -63,7 +63,7 @@ jobs:
           HELM_GCS_CREDENTIALS: ${{ secrets.HELM_GCS_CREDENTIALS }}
           HELM_GCS_REPOSITORY: ${{ secrets.HELM_GCS_REPOSITORY }}
         run: |
-            CHART_VERSION=$(git describe --abbrev=0)
+            CHART_VERSION=$(echo ${GITHUB_REF} | cut -d/ -f 3)
             CHART_NAME=$(awk '/^name:/ {print $2}' ./helm/Chart.yaml)
             export GOOGLE_APPLICATION_CREDENTIALS=${HOME}/helm-gcs-key.json
             echo $HELM_GCS_CREDENTIALS > ${GOOGLE_APPLICATION_CREDENTIALS}


.github/workflows/publish.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           HELM_GCS_CREDENTIALS: ${{ secrets.HELM_GCS_CREDENTIALS }}
           HELM_GCS_REPOSITORY: ${{ secrets.HELM_GCS_REPOSITORY }}
         run: |
-            CHART_VERSION=$(git describe --abbrev=0)
+            CHART_VERSION=$(echo ${GITHUB_REF} | cut -d/ -f 3)
```
kotharironak, line 66: @JBAhire We are still going to cut the release with release-x.y.z tag, right?
In that case, I see we are missing few steps related to that steps and so, we are facing the issue. - https://github.com/hypertrace/attribute-service/blob/main/.circleci/config.yml
The steps were:
- add release tags
- remove trigger tag "release-x.y.z"
- publish charts
- push release tag

kotharironak: As per - https://git-scm.com/docs/git-describe, describes annotated tags. Do the last two tags without `-a` option?

kotharironak: Looks like if you cut the release from Github UI, they are not annotated tag.

JBAhire: we were using `release` prefix for triggering the workflow in CCI. Now in our publish workflow we trigger whenever release is drafted. As we will be adding release note and doing proper release via UI only, I don't find need of release tag anymore.

JBAhire: And about published artifacts. It's same with `release` prefix and without it.

kotharironak (approved):

.github/workflows/publish.yml
jcchavezs: @JBAhire since we are relying on a env var `GITHUB_REF`, I'd add a comment of an example on how that look like or what you expect it to be.

PR #56, opened 2020-12-23
Removes CircleCI Completely

JBAhire:
## Description
As we have added GitHub actions for CI in this repo, removing CircleCI completely. 

GitHub actions handles
- tests on PR
- Build, validate and snyk test on PR
- merge-publish on PR merge
- Publish artifacts and helm charts after creation of release. 

### Testing
- forked-pr: https://github.com/hypertrace/attribute-service/actions/runs/439849456, https://github.com/hypertrace/attribute-service/actions/runs/439849430
- merge-publish: https://github.com/hypertrace/attribute-service/actions/runs/439641204
- Release-publish: https://github.com/hypertrace/attribute-service/actions/runs/439716344 

### Checklist:


kotharironak (dismissed):
lgtm

aaron-steinfeld (dismissed):

aaron-steinfeld (commented):
Hang on - codecov isn't reporting any more.

kotharironak:
There were a couple of things related to codecov here.
a. codecov not showing status on PR (works fine after disabling CCI project)
b. codecov was not able to process uploaded reports - https://codecov.io/gh/hypertrace/attribute-service/commit/c6298427a59e7789e8647ac9bdf9a63efe8632f1

The (b) was the main issue. This was happening due to  report were having old dates and codecov discards such reports - https://community.codecov.io/t/there-was-an-error-processing-coverage-reports-and-ci-not-detected/885
The report file (having date of 6 days ago): (start="1608279207626" 18th dec) 

[jacocoTestReport.xml.txt](https://github.com/hypertrace/attribute-service/files/5738772/jacocoTestReport.xml.txt)

The above seems to be happening due to cache. So, either we can only cache downloaded jars or we can run tests with some extra options. We have tried these options with this draft PR - https://github.com/hypertrace/attribute-service/pull/60, and now it's working fine.

kotharironak (approved):

.github/workflows/pr-test.yml
```diff
@@ -29,33 +29,28 @@ jobs:
             gradle-packages-${{ runner.os }}
 
       - name: Unit test
-        run: ./gradlew jacocoTestReport 
-        
-      - name: Copy unit test reports
-        run: ./gradlew copyAllReports --output-dir=/tmp/test-reports
+        run: ./gradlew -no-build-cache cleanTest jacocoTestReport
```
aaron-steinfeld, line 32:
```suggestion
        run: ./gradlew --no-build-cache jacocoTestReport
```

Any reason for the clean?

Alternatively though, what if we add 
```
codecov:
  max_report_age: off
``` 
to our codecov yaml? That should both let us verify this is the issue in PR, but also allow us to continue to take advantage of caching.

.github/workflows/pr-test.yml
```diff
@@ -29,33 +29,28 @@ jobs:
...
           fail_ci_if_error: true
-          verbose: true
           flags: unit
 
       - name: Integration test
-        run: ./gradlew jacocoIntegrationTestReport 
-
-      - name: Copy integration test reports
-        run: ./gradlew copyAllReports --output-dir=/tmp/test-reports
+        run: ./gradlew -no-build-cache cleanTest jacocoIntegrationTestReport
```
aaron-steinfeld, line 42:
```suggestion
        run: ./gradlew --no-build-cache jacocoIntegrationTestReport
```

.github/workflows/pr-test.yml
JBAhire: we can try that. but let's just go with this for now as I want to use this action: https://github.com/burrunan/gradle-cache-action for handling caching. It has good features and OpenTelemetry also uses it: https://github.com/open-telemetry/opentelemetry-java/blob/master/.github/workflows/pr-build.yml

JBAhire: and also of we resolve this with codecov for now we have to add that codecov config in every repo.

aaron-steinfeld: I'm not sure how the two are related - the linked action looks neat, but it optimizes gradle's cache via the GH one (and adds support for other gradle features like remote caches). Disabling caching on this command would break that action too if we don't change the codecov config.

JBAhire: Ahh. okay. maybe it was not causing any issue for me because it was not using out cache. 

Makes sense to me then. Let's add it here.

aaron-steinfeld (approved):

PR #69, opened 2021-02-03
Update values.yaml

rish691:

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -72,6 +72,10 @@ logConfig:
       enabled: false
 
 config-bootstrapper:
-  job:
-    prefix: attribute
-  configurationCommands: {}
+  jobs:
+    default:
+      enabled: false
+    attribute:
+      prefix: attribute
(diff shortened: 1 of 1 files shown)


helm/values.yaml
```diff
@@ -72,6 +72,10 @@ logConfig:
       enabled: false
 
 config-bootstrapper:
-  job:
-    prefix: attribute
-  configurationCommands: {}
\ No newline at end of file
+  jobs:
```
tim-mwangi, line 75: Do you need to update the config bootstrapper dependency version somewhere?

rish691: Yes, this pr and https://github.com/hypertrace/entity-service/pull/77 needs to be updated with latest config-bootstrapper chart version once 
https://github.com/hypertrace/config-bootstrapper/pull/33 is merged

tim-mwangi: ok.

rish691:
No longer needed

PR #74, opened 2021-02-19
fix: publish test results other than codecov only on non-forked PR

JBAhire:
## Description
- This test results we publish on every commit and PR are helpful but it fails on non-member PR because it requires access to secrets i.e. GitHub Token in this case as we can see here: https://github.com/hypertrace/document-store/runs/1901754228?check_suite_focus=true
- This PR changes trigger for that particular step to run only on non-forked PRs so it will give us data but at the same time won't block PRs from community. 

### Testing
tested this on my own fork: https://github.com/JBAhire/document-store/runs/1933950757?check_suite_focus=true

### Checklist:

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -66,7 +66,7 @@ jobs:
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1.6
-        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           files: ./**/build/test-results/**/*.xml
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -66,7 +66,7 @@ jobs:
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1.6
-        if: always()
+        if: github.event_name == 'push'
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           files: ./**/build/test-results/**/*.xml
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -66,7 +66,7 @@ jobs:
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1.6
-        if: github.event_name == 'push'
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           files: ./**/build/test-results/**/*.xml


.github/workflows/pr-test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1.6
-        if: always()
+        if: github.event_name == 'push'
```
kotharironak, line 69: can we only exclude non-member pull requests? This will exclude member's PR too?

jcchavezs: @kotharironak this is for pushing the test results, not for actually displaying the coverage report isn't @JBAhire ?

JBAhire: @kotharironak this is more of a permissions issue. There's no out of the box way to check it `github.actor` is org member as of now. You can set condition to repository owner but that doesn't solve the problem. There are hacks like [this](https://github.community/t/when-github-actor-is-a-member-not-github-repo-owner/153431/2) which are not worth here considering how small the issue is. 

And @jcchavezs , this is for displaying comment with test summary like this: https://github.com/hypertrace/document-store/pull/35#issuecomment-779141554 . 

This use-case is also satisfied by codecov upto some extent now so I won't bother enough about it as of now.

kotharironak: I was referring to `github.event.pull_request.head.repo.full_name != github.repository` to check if this is forked PR

JBAhire: yup. updated

.github/workflows/pr-test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1.6
-        if: always()
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```
kotharironak, line 69: we don't need this now - github.event_name == 'push'

kotharironak: Got it, we have 
```
	on:
  push:	  push:
    branches:	    branches:
      - main	      - main
  pull_request:	  pull_request:
```

kotharironak (approved):

.github/workflows/pr-test.yml
JBAhire: it it to run this step on push event along with pull_request event.

.github/workflows/pr-test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
      
       - name: Publish Unit Test Results
         uses: docker://ghcr.io/enricomi/publish-unit-test-result-action:v1.6
-        if: always()
+        if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
```
aaron-steinfeld, line 69: Will this still run regardless of the failure of a previous step (i.e. if the tests fail)? That's the behavior we want, I'm just not sure if github treats `always()` as special - I guess we'll see

kotharironak: @JBAhire we can test this by introducing failure in the test and creating dummy PR.

PR #82, opened 2021-03-31
feat: add way of defining an attribute from first class field

aaron-steinfeld:
## Description
We previously allowed defining an attribute based on its string key, but certain fields in a span and trace are treated as first class and only accessible through those concrete fields. This change adds support to the attribute definition schema to access those values.

### Testing
Schema change only (but used this functionality as part of an e2e verification)

### Checklist:
- [N/A] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
+++ b/attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
@@ -144,6 +144,13 @@ message AttributeDefinition {
   oneof value {
     Projection projection = 1;
     string source_path = 2;
+    SourceField source_field = 3;
+  }
+
+  enum SourceField {
+    SOURCE_FIELD_UNSET = 0;
+    SOURCE_FIELD_START_TIME = 1;
+    SOURCE_FIELD_END_TIME = 2;
   }
 }
 


attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -144,6 +144,13 @@ message AttributeDefinition {
   oneof value {
     Projection projection = 1;
     string source_path = 2;
+    SourceField source_field = 3;
+  }
+
+  enum SourceField {
```
tim-mwangi, line 150: This cannot be a new projection definition?

aaron-steinfeld: all attribute definitions eventually resolve down to data - so for example, an attribute like `span.startTime` needs to be able to specify it's `source_field = SOURCE_FIELD_START_TIME` while something like `span.requestUri` could already do `source_path = "http.request.uri"` (or whatever tag that would correspond to). The projections come in when we go a level up. So for example, let's say we introduce an alias for `span.startTime` called `span.timestamp` - that'd use an attribute_id type projection. Does that make sense?

aaron-steinfeld:
Updated deps first in https://github.com/hypertrace/attribute-service/pull/83

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
tim-mwangi: By a level up you mean like in Pinot?

aaron-steinfeld: not necessarily. When we read attributes using the attribute reader ( https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java ), it uses the definition and recursively resolves it. So taking parts of the example before, we could imagine something like (pseudocode for brevity):
```
{
  id: span.startTime,
  definition: {
    source_field = SOURCE_FIELD_START_TIME
  }
},
{
  id: span.endTime,
  definition: {
    source_field = SOURCE_FIELD_END_TIME
  }
},
{
  id: span.duration,
  definition: {
    projection: {
      expression: {
        operator: MINUS,
        arguments: [
           {
              attribute_id: span.startTime,
              attribute_id: span.endTime
           }
        ]
      }
   }
}
```

To resolve a projection, we have to first resolve its arguments - walking down the tree until we get to the leaves which are either literals, source_path (that is, tag keys), or the new feature added here, source_field - that is, first class fields on a span/trace. Without that resolution, any projections could not use values from fields like startTime, endTime, spanId, traceId etc.

aaron-steinfeld: As to the pinot question, query service does not read source_path or source_field, it only reads projections (since those can be rewritten as new queries) .

tim-mwangi: Ok. Totally overlooked source_path. Makes sense.

tim-mwangi (approved):

PR #89, opened 2021-04-15
feat: Log event attribute scope

rish691:


aaron-steinfeld (changes requested):
This entire enum is deprecated, scopes are no longer hard coded.

rish691:
> This entire enum is deprecated, scopes are no longer hard coded.

Ohk, so I guess adding attributes in config-bootstrapper should suffice.

aaron-steinfeld:
> > This entire enum is deprecated, scopes are no longer hard coded.
> 
> Ohk, so I guess adding attributes in config-bootstrapper should suffice.

yep - be sure to use the `scope_string` field rather than `scope` (at least, I think that's the name - hopefully there are examples!)

PR #99, opened 2021-08-23
Unary api

aaron-steinfeld:
## Description
Streaming api introduces more complexity, and using them incorrectly we suspect is leading to memory leaks. Here we 

### Testing
Added and updated UTs

### Checklist:

### DIFF
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
@@ -1,5 +1,8 @@
 package org.hypertrace.core.attribute.service;
 
+import static java.util.Objects.isNull;
+
+import com.google.common.collect.Streams;
 import com.google.protobuf.ServiceException;
 import com.typesafe.config.Config;
 import io.grpc.stub.StreamObserver;
@@ -27,6 +30,8 @@ import org.hypertrace.core.attribute.service.v1.AttributeSource;
 import org.hypertrace.core.attribute.service.v1.AttributeSourceMetadataDeleteRequest;
 import org.hypertrace.core.attribute.service.v1.AttributeSourceMetadataUpdateRequest;
 import org.hypertrace.core.attribute.service.v1.Empty;
+import org.hypertrace.core.attribute.service.v1.GetAttributesRequest;
+import org.hypertrace.core.attribute.service.v1.GetAttributesResponse;
 import org.hypertrace.core.documentstore.Collection;
 import org.hypertrace.core.documentstore.Datastore;
 import org.hypertrace.core.documentstore.DatastoreProvider;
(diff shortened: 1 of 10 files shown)


attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -310,26 +315,49 @@ public void findAll(Empty request, StreamObserver<AttributeMetadata> responseObs
     }
   }
 
+  @Override
+  public void getAttributes(
+      GetAttributesRequest request, StreamObserver<GetAttributesResponse> responseObserver) {
+    String tenantId = RequestContext.CURRENT.get().getTenantId().orElse(null);
```
jcchavezs, line 321: Maybe this is never going to happen but no nil checks needed here? Also, this looks a bit opposite to [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
```diff
@@ -295,6 +296,97 @@ public void testFindAttributes() {
     }
   }
 
+  @Test
+  public void testGetAttributes() {
```
jcchavezs, line 300: I think this test is quite long and I think it is because the objects' creation is very long. I am not sure every single attribute in the objects is relevant for the test, How about abstracting them in private statics and make this test focus on what matters?

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
aaron-steinfeld: No nil checks needed, this is a method invoked by GRPC and the args can never be null. As for the tenant ID, I tend to agree, but it's debatable and this is one of several methods in the service so I valued consistency more.

My preferred style is to just pull the request context and pass it on to the downstream code. I'd guess the counterarguments that would be raised here would look like:
- fail fast if the request is missing a tenant id. IMO this is meaningful - we don't want to do a bunch of work that doesn't require the tenant id before realizing we can't complete (assuming we don't have transactionality, which would make this even more important)
- simplify the downstream code so it doesn't need its own checks everywhere (this is kind of the whole point though, so this argument I don't buy)

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
aaron-steinfeld: Yeah, since this is a new api over an existing set of functionality, I didn't write this test from scratch. I took the test from the old api (which was significantly _more_ verbose, if you can imagine) and used it again here to make sure they behaved consistently. I can refactor those down a bit.

skjindal93 (approved):

GurtejSohi (approved):

PR #100, opened 2021-08-25
Add Shutdown and grpc health check

aaron-steinfeld:
## Description
This combines two different changes in an attempt to be a more well behaved server. It cleans up (or actually adds, since this server was missing it entirely) server shutdown, and it adds a real grpc service as a health check as described at https://github.com/grpc/grpc/blob/master/doc/health-checking.md

The idea is to get a check that better reflects the actual health of the service, as the current check continues to serve OK when the grpc server has gone non-responsive. It also appropriately waits to mark healthy until the server is actually ready, and offline _before_ shutdown, hopefully leading to smoother rollouts/upgrades.

Depends on https://github.com/hypertrace/docker-java-base/pull/9

### Testing
Built and tested this change locally in docker

### DIFF
--- a/attribute-service/build.gradle.kts
+++ b/attribute-service/build.gradle.kts
@@ -57,8 +57,9 @@ dependencies {
   implementation(project(":attribute-service-impl"))
 
   implementation("org.hypertrace.core.serviceframework:platform-service-framework:0.1.28")
-  implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.5.2")
+  implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.6.0")
   implementation("org.hypertrace.core.documentstore:document-store:0.5.4")
+  implementation("io.grpc:grpc-services:1.40.0")
 
   // Logging
   implementation("org.slf4j:slf4j-api:1.7.30")
@@ -102,7 +103,9 @@ tasks.jacocoIntegrationTestReport {
 hypertraceDocker {
   defaultImage {
     javaApplication {
-      port.set(9012)
+      ports.add(9012)
+      adminPort.set(9013)
+      healthCheck.set("HEALTHCHECK --interval=2s --start-period=15s --timeout=2s CMD /bin/grpc_health_probe -addr=:9012 -connect-timeout=100ms -rpc-timeout=150ms")
     }
   }
 }
--- a/attribute-service/build.gradle.kts
+++ b/attribute-service/build.gradle.kts
@@ -57,7 +57,7 @@ dependencies {
   implementation(project(":attribute-service-impl"))
 
   implementation("org.hypertrace.core.serviceframework:platform-service-framework:0.1.28")
-  implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.4.0")
+  implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.5.2")
   implementation("org.hypertrace.core.documentstore:document-store:0.5.4")
 
   // Logging
@@ -65,7 +65,7 @@ dependencies {
   runtimeOnly("org.apache.logging.log4j:log4j-slf4j-impl:2.14.1")
 
   // GRPC
-  runtimeOnly("io.grpc:grpc-netty:1.36.1")
+  runtimeOnly("io.grpc:grpc-netty:1.40.0")
   constraints {
     runtimeOnly("io.netty:netty-codec-http2:4.1.61.Final") {
       because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1089809")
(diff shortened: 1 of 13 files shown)


attribute-service/build.gradle.kts
```diff
@@ -102,7 +103,9 @@ tasks.jacocoIntegrationTestReport {
 hypertraceDocker {
   defaultImage {
     javaApplication {
-      port.set(9012)
+      ports.add(9012)
+      adminPort.set(9013)
+      healthCheck.set("HEALTHCHECK --interval=2s --start-period=15s --timeout=2s CMD /bin/grpc_health_probe -addr=:9012 -connect-timeout=100ms -rpc-timeout=150ms")
```
aaron-steinfeld, line 108: don't love this, but our base plugin doesn't and shouldn't assume a grpc server for its default health check.

attribute-service/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceEntry.java
```diff
@@ -48,11 +53,14 @@ protected void doStart() {
...
-  protected void doStop() {}
+  protected void doStop() {
+    healthStatusManager.enterTerminalState();
+    ServerManagementUtil.shutdownServer(this.server, this.getServiceName(), Duration.ofMinutes(1));
+  }
 
   @Override
   public boolean healthCheck() {
-    return true;
+    throw new UnsupportedOperationException("Please use grpc health check");
```
aaron-steinfeld, line 63: Alternatively, I can try to serve this one up, too. Thee issue is that health status manager only exposes write apis, I see no way of reading the value without a full grpc client.

aaron-steinfeld: ah, integration tests use it. thought this might backfire. So I can make that path work, or I can try to pipe the grpc result into here...

avinashkolluru: > pipe the grpc result into here.

This would be ideal in my opinion

helm/templates/deployment.yaml
```diff
@@ -75,13 +72,12 @@ spec:
           livenessProbe:
             initialDelaySeconds: {{ int .Values.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ int .Values.livenessProbe.periodSeconds }}
-            tcpSocket:
-              port: grpc-port
+            exec:
+              command: [ "/bin/grpc_health_probe", "-addr=:{{ .Values.containerHealthProbePort }}", "-connect-timeout=100ms", "-rpc-timeout=150ms" ]
```
avinashkolluru, line 76: How is this `/bin/grpc_health_probe` getting bundled?

avinashkolluru: As per https://github.com/grpc-ecosystem/grpc-health-probe#example-grpc-health-checking-on-kubernetes dont we need to add this into our Dockerfile

```
RUN GRPC_HEALTH_PROBE_VERSION=v0.3.1 && \
    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
    chmod +x /bin/grpc_health_probe
```

skjindal93: I think it's coming from https://github.com/hypertrace/docker-java-base/pull/9

aaron-steinfeld: See the linked PR - hypertrace/docker-java-base#9 

But I'm wondering if we even need that, and can make this more low touch. If I pipe the grpc result in here but leave the existing checks (just converting that socket check to a get - if there's not a story there), wondering if I actually miss out on anything.

helm/templates/deployment.yaml
```diff
@@ -75,8 +75,9 @@ spec:
           livenessProbe:
             initialDelaySeconds: {{ int .Values.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ int .Values.livenessProbe.periodSeconds }}
-            tcpSocket:
```
aaron-steinfeld, line 78: @ravisingal  (or anyone else) - do you have context on why the liveness probe was a tcp socket but the readiness probe was a health get?

helm/templates/deployment.yaml
avinashkolluru: Sorry. Missed the linked PR hypertrace/docker-java-base#9

avinashkolluru: >  If I pipe the grpc result in here but leave the existing checks (just converting that socket check to a get - if there's not a story there), wondering if I actually miss out on anything.

Good point. In that case the change in base image to include the grpc_health_probe can be avoided.

helm/templates/deployment.yaml
aaron-steinfeld: (he did not - one of those mysteries!)

laxmanchekka (approved):

avinashkolluru (dismissed):

aaron-steinfeld:
Will pull in `https://github.com/hypertrace/java-grpc-utils/pull/21` first

avinashkolluru (approved):

PR #105, opened 2021-09-02
Internal flag in attribute_metadata

ankitchoudhary111:
This flag will tell whether the attribute is meant for internal use only or it can be used externally as well (i.e. can be exposed to the UI).

### DIFF
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
@@ -251,8 +251,8 @@ public class AttributeMetadataModel implements Document {
     return isInternal;
   }
 
-  public void setIsInternal(boolean is_internal) {
-    this.isInternal = is_internal;
+  public void setIsInternal(boolean isInternal) {
+    this.isInternal = isInternal;
   }
 
   public AttributeMetadata toDTO() {
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
@@ -343,7 +343,7 @@ public class AttributeMetadataModel implements Document {
         + sources
         + ", metadata="
         + metadata
-        + ", Internal="
+        + ", internal="
         + internal
         + '}';
   }
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
@@ -74,8 +74,7 @@ public class AttributeMetadataModel implements Document {
   @JsonDeserialize(using = AttributeDefinitionDeserializer.class)
   private AttributeDefinition definition = AttributeDefinition.getDefaultInstance();
 
-  @JsonProperty(value = "is_internal")
-  private boolean isInternal;
+  private boolean internal;
 
   protected AttributeMetadataModel() {}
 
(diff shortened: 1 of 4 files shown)


attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -243,6 +247,14 @@ public void setDefinition(AttributeDefinition definition) {
     this.definition = definition;
   }
 
+  public boolean getIsInternal() {
+    return isInternal;
+  }
+
+  public void setIsInternal(boolean is_internal) {
+    this.isInternal = is_internal;
+  }
```
skjindal93, line 256:
```suggestion
  public void setIsInternal(boolean isInternal) {
    this.isInternal = isInternal;
  }
```

skjindal93 (commented):
Can you please add a description in the PR, explaining the need for the internal flag?

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -243,6 +246,14 @@ public void setDefinition(AttributeDefinition definition) {
     this.definition = definition;
   }
 
+  public boolean getInternal() {
+    return internal;
+  }
+
+  public void setInternal(boolean isInternal) {
+    this.internal = isInternal;
+  }
```
skjindal93, line 255: since, we changed the field to `internal` now
```suggestion
  public void setInternal(boolean internal) {
    this.internal = internal;
  }
```

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -331,6 +343,8 @@ public String toString() {
         + sources
         + ", metadata="
         + metadata
+        + ", Internal="
```
skjindal93, line 346:
```suggestion
        + ", internal="
```

skjindal93:
I believe there are integration tests for attribute service. Can you please add/modify the integration tests?

attribute-service/src/integrationTest/java/org/hypertrace/core/attribute/service/AttributeServiceTest.java
```diff
@@ -126,6 +128,40 @@ private void testCreateInvalid(boolean useRequestHeaders) {
...
+
+    AttributeCreateRequest request =
+        AttributeCreateRequest.newBuilder().addAttributes(expectedAttributeMetadata).build();
+    client.create(requestHeaders, request);
+
+    List<AttributeMetadata> attributeMetadataList =
+        Streams.stream(
+                client.findAttributes(requestHeaders, AttributeMetadataFilter.getDefaultInstance()))
+            .collect(Collectors.toList());
+    assertEquals(expectedAttributeMetadata, attributeMetadataList.get(0));
```
skjindal93, line 162: nit: if the following works
```suggestion
    assertEquals(List.of(expectedAttributeMetadata), attributeMetadataList);
```

else, please assert the size of the `attributeMetadataList` as well

skjindal93 (approved):

attribute-service/src/integrationTest/java/org/hypertrace/core/attribute/service/AttributeServiceTest.java
```diff
@@ -126,6 +128,40 @@ private void testCreateInvalid(boolean useRequestHeaders) {
...
+            .setInternal(true)
+            .build();
+
+    AttributeCreateRequest request =
+        AttributeCreateRequest.newBuilder().addAttributes(expectedAttributeMetadata).build();
+    client.create(requestHeaders, request);
+
+    List<AttributeMetadata> attributeMetadataList =
+        Streams.stream(
+                client.findAttributes(requestHeaders, AttributeMetadataFilter.getDefaultInstance()))
```
aaron-steinfeld, line 160: hmm we need to deprecate this - didn't realize we still used this client. anyway, out of scope here.

aaron-steinfeld (approved):
can you update the snyk ignore expiration either in this PR or a separate one?

PR #106, opened 2021-09-16
added internal flag to the AttributeMetadataFilter

ankitchoudhary111:


aaron-steinfeld:
Not sure if this was meant for review yet or not, but the implementation is missing (see https://github.com/hypertrace/attribute-service/blob/main/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java#L361 ) and tests please.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -385,6 +388,8 @@ private Query getQueryForFilter(
       andFilters.add(new Filter(Filter.Op.IN, ATTRIBUTE_KEY_KEY, keyFilterRequest));
     }
 
+    andFilters.add(new Filter(Op.IN, ATTRIBUTE_INTERNAL_KEY, internalFilter));
```
avinashkolluru, line 391: Integration test for this?

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -367,6 +369,7 @@ private Query getQueryForFilter(
             .collect(Collectors.toUnmodifiableList());
     List<String> keyFilterRequest = attributeMetadataFilter.getKeyList();
     List<String> fqnFilterRequest = attributeMetadataFilter.getFqnList();
+    boolean internalFilter = attributeMetadataFilter.getInternal();
```
skjindal93, line 372:
```suggestion
    boolean isInternal = attributeMetadataFilter.getInternal();
```

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
skjindal93:
```suggestion
    andFilters.add(new Filter(Op.EQ, ATTRIBUTE_INTERNAL_KEY, internalFilter));
```

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -136,6 +136,7 @@ message AttributeMetadataFilter {
   repeated AttributeScope scope = 2 [deprecated = true]; // Use scope_string instead
   repeated string key = 4;
   repeated string scope_string = 5;
+  bool internal = 6;
```
skjindal93, line 139: discussed offline to have this as

```suggestion
  optional bool internal = 6;
```

because, for the default filter instance, we would want all the attributes, and not just the external attributes

aaron-steinfeld:
LGTM - will let @skjindal93 re-review.

For snyk, can we please get that green either in this PR or as a followup.

skjindal93 (approved):

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
```diff
@@ -200,14 +201,18 @@ public void testFindAttributes() {
 
       Filter filter = queryCaptor.getValue().getFilter();
       // The structure of the filters is an and(^) filter chain that looks like this:
-      // (((tenant_id ^ fqn) ^ (scope_string | scope)) ^ key)
+      // ((((tenant_id ^ fqn) ^ (scope_string | scope)) ^ key) ^ internal)
```
jcchavezs, line 204: The test is monster big, anyways I wonder if we test both scenarios where internal is present and when it isn't (as it is optional).

ankitchoudhary111: I have added a separate test for the scenario when internal is not present.

skjindal93 (approved):

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
jcchavezs: I am curious where exactly

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
```diff
@@ -150,7 +150,119 @@ public void testFindAllNoTenantId() {
...
+
+      verify(responseObserver, times(1)).onCompleted();
+      verify(responseObserver, never()).onError(any(Throwable.class));
+    } finally {
+      ctx.detach(previous);
+    }
+  }
+
+  @Test
+  public void testFindAttributesWithNoInternalFlag() {
```
ankitchoudhary111, line 265: @jcchavezs for case with no internal

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
```diff
@@ -150,7 +150,119 @@ public void testFindAllNoTenantId() {
   }
 
   @Test
-  public void testFindAttributes() {
+  public void testFindAttributesWithInternalFlag() {
```
ankitchoudhary111, line 153: @jcchavezs for case where internal is present.

aaron-steinfeld (approved):

PR #111, opened 2021-11-12
chore : updated config bootstrapper

sarthak77:
Updated config bootstrapper version and grpc versions.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -24,4 +24,4 @@ version: 0.1.0
 dependencies:
   - name: config-bootstrapper
     repository: "https://storage.googleapis.com/hypertrace-helm-charts"
-    version: 0.2.20
+    version: 0.2.21
--- a/.snyk
+++ b/.snyk
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2021-10-31T00:00:00.000Z
+        expires: 2021-12-31T00:00:00.000Z
 patch: {}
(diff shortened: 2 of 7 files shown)


helm/Chart.yaml
```diff
@@ -24,4 +24,4 @@ version: 0.1.0
 dependencies:
   - name: config-bootstrapper
     repository: "https://storage.googleapis.com/hypertrace-helm-charts"
-    version: 0.2.20
+    version: 0.2.21
```
kotharironak, line 27: I know this is required only for attribute service, should we also update at Entity Service too, just to be on the same version of config-bootstraper on both the hook - https://github.com/hypertrace/entity-service/blob/main/helm/Chart.yaml#L27

sarthak77: CC @aaron-steinfeld

aaron-steinfeld: It makes sense. We know there's no reason to and it won't matter since we saw the config bootstrapper change - but aligning them will help the next person who doesn't have the context.

aaron-steinfeld: This repo is failing snyk though, can we fix that - either in this PR or a separate one. Easiest fix will be to upgrade grpc and grpc-utils, we fixed this vulnerability in the utils (and we want to keep the grpc version aligned with the utils).

kotharironak: let's update as part of this PR.

sarthak77: updated

kotharironak:
To fix snyk, change the .snyk file for expiry data - https://github.com/hypertrace/attribute-service/blob/8baa62e9460458fa418ceef65c69edb3c015282e/.snyk#L8

Change to Dec end - `2021-10-31T00:00:00.000Z` - >`2021-12-31T00:00:00.000Z`

kotharironak (approved):

GurtejSohi (approved):

PR #119, opened 2022-02-04
chore: support attribute category

skjindal93:
## Description
Attributes can be categorized under various buckets, like Performance/Infrastructure

### Testing
Added unit tests

### Checklist:


aaron-steinfeld (commented):
Discussed offline, not sure this is broadly applicable to attributes everywhere. Until we have a compelling reason for them, would rather not merge.

PR #123, opened 2022-05-06
added new scope

japoorv:
Added new `SENSITIVE_SPAN` scope for sensitive data attributes.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,4 +1,5 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
 version: v1.14.1
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore: {}
 patch: {}
--- a/.snyk
+++ b/.snyk
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2022-04-30T00:00:00.000Z
+        expires: 2022-05-30T00:00:00.000Z
 patch: {}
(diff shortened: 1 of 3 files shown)


avinashkolluru (approved):

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -28,6 +28,7 @@ enum AttributeScope {
   CLUSTERS_SNAPSHOT = 17;
   DOMAIN = 18;
   ACTOR = 19;
+  SPAN_SENSITIVE_DATA = 20;
```
skjindal93, line 31: Can we rename it to `SENSITIVE_SPAN` or `SENSITIVE_EVENT`? We interchangeably use `SPAN` and `EVENT` in attribute scopes

avinashkolluru: No EVENT please. EVENT was wrongly named. It should have been SPAN. So lets not continue to name new enums that way.

skjindal93 (approved):

.snyk
```diff
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2022-04-30T00:00:00.000Z
+        expires: 2022-05-30T00:00:00.000Z
```
avinashkolluru, line 8: Isn't this vulnerability fixed with the upgrade?

japoorv: The initial issue was the expiration one, the other one was due to netty issue.
Upgraded the version, everything fine now

japoorv: Removing this now

avinashkolluru (approved):

PR #150, opened 2022-11-07
Introduce the read-time "custom" flag with the ability to filter by the same

suresh-prakash:


attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -50,6 +50,7 @@ enum AttributeType {
   TYPE_UNDEFINED = 0;
   ATTRIBUTE = 1;
   METRIC = 2;
+  CUSTOM_ATTRIBUTE = 3;
```
aaron-steinfeld, line 53: Would not change this. Let's discuss what you're trying to accomplish.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -285,7 +286,8 @@ public AttributeMetadata.Builder toDTOBuilder() {
                                     .putAllSourceMetadata(stringMapEntry.getValue())
                                     .build())))
             .setDefinition(this.definition)
-            .setInternal(internal);
+            .setInternal(internal)
+            .setCustom(tenantId != null && !ROOT_TENANT_ID.equals(tenantId));
```
aaron-steinfeld, line 290: nit: can tenantId ever be null? Hopefully this is handled on the way in

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
```diff
@@ -459,7 +513,13 @@ private Document createMockDocument(
         .append(scope.name())
         .append(" ")
         .append(key)
-        .append("\"}");
+        .append("\"");
+
+    if (tenantId != null && !ROOT_TENANT_ID.equals(tenantId)) {
```
aaron-steinfeld, line 518: this wouldn't be persisted right? Hopefully test works without this.

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModelTest.java
```diff
@@ -353,4 +355,151 @@ void testScopeStringCompatibility() throws IOException {
...
+            .setScope(AttributeScope.EVENT)
+            .setScopeString(AttributeScope.EVENT.name())
+            .setType(AttributeType.ATTRIBUTE)
+            .setUnit("ms")
+            .setValueKind(AttributeKind.TYPE_STRING)
+            .setDefinition(
+                AttributeDefinition.newBuilder()
+                    .setProjection(Projection.newBuilder().setAttributeId("test"))
+                    .build())
+            .setInternal(true)
```
aaron-steinfeld, line 399: I don't think this test is a valid case, but if we did want to keep it - setCustom(false) would be appropriate here. It's equivalent to the default boolean value which is why the test is passing, but omitting it in one test and not the other indicates to the reader we expect different behavior.

attribute-service/src/integrationTest/java/org/hypertrace/core/attribute/service/AttributeServiceTest.java
```diff
@@ -76,15 +80,24 @@ public class AttributeServiceTest {
...
 
   private static AttributeServiceClient client;
+  private static AttributeServiceBlockingStub stub;
 
   @BeforeAll
   public static void setup() {
     System.out.println("Testing the Attribute E2E Test");
     IntegrationTestServerUtil.startServices(new String[] {"attribute-service"});
     Channel channel = ManagedChannelBuilder.forAddress("localhost", 9012).usePlaintext().build();
+    //    Channel channel2 = ManagedChannelBuilder.forAddress("localhost",
```
aaron-steinfeld, line 93: Please clean up this test

attribute-service/src/integrationTest/java/org/hypertrace/core/attribute/service/AttributeServiceTest.java
```diff
@@ -76,15 +80,19 @@ public class AttributeServiceTest {
           .build();
 
   private final Map<String, String> requestHeaders = Map.of("x-tenant-id", TEST_TENANT_ID);
+  private final Map<String, String> systemRequestHeaders = Map.of("x-tenant-id", ROOT_TENANT_ID);
 
-  private static AttributeServiceClient client;
+  private static AttributeServiceBlockingStub stub;
```
suresh-prakash, line 85: Changes in this file:
1. Replace `client` with `blockingStub` since the `getAttributes()` API is not present in the `client`
2. Update the tests to test the `getAttributes()` API instead of the deprecated `findAttributes()` API
3. Add new tests to test without custom filtering and with custom filtering set to true/false.

skjindal93 (dismissed):

attribute-service/src/integrationTest/java/org/hypertrace/core/attribute/service/AttributeServiceTest.java
suresh-prakash: Makes sense.
No new integration test is needed since we are not adding the ability to filter by "custom".

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -185,9 +186,10 @@ public void delete(AttributeMetadataFilter request, StreamObserver<Empty> respon
       return;
     }
 
-    Query query = new Query();
-    query.setFilter(getTenantIdEqFilter(tenantId.get()));
-    Iterator<Document> documents = collection.search(getQueryForFilter(tenantId.get(), request));
+    // Make sure to delete only the custom attributes
+    final AttributeMetadataFilter updatedRequest = request.toBuilder().setCustom(true).build();
```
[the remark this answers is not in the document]
suresh-prakash: Talked to Shubham and it seems that UI can do the filtering since it anyways query for the whole attribute set.

attribute-service/src/integrationTest/java/org/hypertrace/core/attribute/service/AttributeServiceTest.java
```diff
@@ -94,7 +102,10 @@ public static void teardown() {
 
   @BeforeEach
   public void setupMethod() {
-    client.delete(requestHeaders, AttributeMetadataFilter.getDefaultInstance());
+    executeWithHeadersContext(
```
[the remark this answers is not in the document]
suresh-prakash: Reverted changes.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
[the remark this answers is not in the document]
suresh-prakash: > Do we ever query just custom attributes - or just system attributes (i.e. is there a use case for that)?

Also, will be useful to rate-limit the number of custom attributes created.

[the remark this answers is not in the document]
aaron-steinfeld: Yes, deleting system would be done from that context - in other words, we don't have hierarchy on delete. You only delete for your tenant, so the filter shouldn't be required there

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeMetadataValidator.java
```diff
@@ -57,6 +59,14 @@ public static void validate(AttributeCreateRequest attributeCreateRequest) {
     }
   }
 
+  public static void validateDeletionFilter(final AttributeMetadataFilter attributeMetadataFilter) {
+    if (!attributeMetadataFilter.getCustom()) {
```
aaron-steinfeld, line 63: This would mean all places that currently call delete would fail validation, since false is the ddefault value. We'd need to do a presence check (`hasCustom() && !getCustom()`) (and would need to make sure we're appropriately defaulting to true in the delete case, false in the query case)

suresh-prakash: As covered by the new integration tests.

aaron-steinfeld: > This is not required, right? Because, in the query case, we have 3 different meaning

Yes, sorry - my mixup. I meant defaulting to not treat it like false, but it shouldn't be true - it's as you described.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -180,16 +180,18 @@ public void updateSourceMetadata(
   }
 
   @Override
-  public void delete(AttributeMetadataFilter request, StreamObserver<Empty> responseObserver) {
-    validateDeletionFilter(request);
+  public void delete(
+      final AttributeMetadataFilter request, final StreamObserver<Empty> responseObserver) {
+    final AttributeMetadataFilter modifiedRequest = validateAndUpdateDeletionFilter(request);
```
aaron-steinfeld, line 185: Had a similar comment on one of the other PRs, but if this throws, nothing is catching it and reporting it to the osberver.

suresh-prakash: Good catch. Handled now.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -190,40 +191,50 @@ public void delete(
...
-                    AttributeMetadata metadata =
-                        AttributeMetadataModel.fromJson(attrMetadataDoc).toDTO();
-                    boolean response =
-                        collection.delete(AttributeMetadataDocKey.from(tenantId.get(), metadata));
-                    if (!response) {
+    try {
+      Iterator<Document> documents =
+          collection.search(getQueryForFilter(tenantId.get(), modifiedRequest));
+      boolean status =
+          StreamSupport.stream(Spliterators.spliteratorUnknownSize(documents, 0), false)
```
aaron-steinfeld, line 198: nit -` Streams.stream(documents)` (since we already are using guava)

Edit: realized this just got wrapped and was pre-existing, don't need to fix

aaron-steinfeld (approved):

PR #152, opened 2022-11-17
Introduced add and delete methods in the caching client with invalidation

suresh-prakash:


caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/CachingAttributeClient.java
```diff
@@ -22,6 +24,10 @@ public interface CachingAttributeClient {
 
   Single<List<AttributeMetadata>> getAllInScope(String scope);
 
+  Completable create(final Collection<AttributeMetadata> attributeMetadata);
```
aaron-steinfeld, line 27: As I think this through more, one thing we should consider is that attributes are read by many services in the platform. So imagine the following flow:

1. A user creates (or deletes) an attribute with an API call (or eventually UI)
2. We use this client in GQL to make that change both to local state and call attribute service to persist to disk
3. The updated collection is returned
4. Now, another call is made - all we know is attribute service and that one replica are up to date. If that call goes to a different GQL replica, it could get stale results (if an attribute query) or an error (if trying to query that attribute). Same goes even in a single replica deployment, since gateway too has an attribute cache.

So the system would eventually become consistent - I think the timeouts are maybe 15m? But it's fairly typical to try out something right after you create it. The only options I really see here are 
1. accepting the eventual consistency (potentially with a significant decrease on the expiration for attributes, say to 1m)
2. introducing cache invalidation events that can be broadcast

Do you see any other approaches?

aaron-steinfeld (commented):
other than the consistency discussion change looks good.

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/CachingAttributeClient.java
suresh-prakash: Right. That makes sense and I don't see any other approach.
I think for our use-case, it makes sense to accept eventual consistency. We could always implement cache-invalidation events later when there's a need. Updated the default cache invalidation interval to 1 minute. All clients (unless they override) will use this cache-invalidation interval. Note this means, we'll have to make changes to the clients till GraphQL.

Having said that, I feel, it doesn't stop us from implementing `create` and `delete` in the caching client as it would benefit
1. any service with single-replica deployment
2. subsequent retries could return the latest results (even before cache expiration) in the case of multi-replica deployed services

The downside of this being, there could be inconsistent results across different invocations (as each invocation could hit a different replica) until cache expiration. But, this effect is unavoidable even in the case of accepting eventual consistency since different replicas might expire the cache at different times (though all of them have a constant 1 minute cache invalidation interval).

@skjindal93 @aaron-steinfeld What do you think?

aaron-steinfeld: I think this is a sensible first cut. I'm not sure what the reasonable default cache expiration is - tradeoff is that when we do a mutation, we'll want that asap - but that's likely very infrequent (a couple times in the lifetime of a tenant), vs fetching the attributes with the shorter expiration will basically mean we have that many more calls per minute always. Maybe 3m? 

Also curious if this cache is async reloading or not - probably worth updating if it's not, because a short expiration will lead to a lot more inline calls.

aaron-steinfeld: > Also curious if this cache is async reloading or not - probably worth updating if it's not, because a short expiration will lead to a lot more inline calls.

It's not, but can disregard for now. It would be a bigger change right now because the cache is lazy

skjindal93 (approved):

aaron-steinfeld (approved):

PR #153, opened 2022-11-17
Limit the number of custom attributes that can be created

suresh-prakash:


attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -92,25 +98,33 @@ private Datastore initDataStore(Config config) {
...
-    if (tenantId.isEmpty()) {
+    Optional<String> tenantIdOptional = RequestContext.CURRENT.get().getTenantId();
+    if (tenantIdOptional.isEmpty()) {
       responseObserver.onError(new ServiceException("Tenant id is missing in the request."));
       return;
     }
 
+    final String tenantId = tenantIdOptional.orElseThrow();
+
     AttributeMetadataValidator.validate(request);
```
aaron-steinfeld, line 109: Not sure if this is from your other change or pre-existing but this looks like a bug to me. There's no try/catch around this is there? A validation error would fail but never be reported to the caller who would just hang. We probably should wrap the whole create body in a try/catch rather than the individual bits being done.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -339,6 +353,26 @@ public void getAttributes(
     responseObserver.onCompleted();
   }
 
+  private void verifyCustomAttributeLimitReached(final String tenantId, final int newAttributeCount)
+      throws StatusException {
+    if (ROOT_TENANT_ID.equals(tenantId)) {
+      // No rate limit validation for system attributes
```
aaron-steinfeld, line 359: nit - it's not really a rate (there's no time aspect), just a limit

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
```diff
@@ -92,25 +98,33 @@ private Datastore initDataStore(Config config) {
...
           new AttributeMetadataDocKey(
-              tenantId.get(),
-              attributeMetadataModel.getScopeString(),
-              attributeMetadataModel.getKey()),
+              tenantId, attributeMetadataModel.getScopeString(), attributeMetadataModel.getKey()),
           attributeMetadataModel);
     }
+
+    try {
+      verifyCustomAttributeLimitReached(tenantId, attributeDocs.size());
```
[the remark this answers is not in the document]
suresh-prakash: > this is really another form of validation, does it belong in the validator?

Yes. Moved it to the validator class with some refactoring.

> Also name is inverted - we want to make sure the limit hasn't been reached.

Right. Updated.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeServiceImpl.java
suresh-prakash: Good catch. Added try-catch block.

aaron-steinfeld (approved):

PR #154, opened 2022-11-28
Add the "value_externally_writable" flag to attribute metadata

suresh-prakash:


attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -103,6 +103,8 @@ message AttributeMetadata {
   string scope_string = 17;
   bool internal = 18;
   bool custom = 19;
+  // Whether the attribute value can be created, updated or deleted externally
```
aaron-steinfeld, line 106: how is this different than the two existing flags - custom and internal?

suresh-prakash:
Not required.

PR #155, opened 2022-12-05
Attribute metadata update implementation

suresh-prakash:


attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_service.proto
```diff
@@ -19,6 +19,7 @@ service AttributeService {
     option deprecated = true;
   };
   rpc GetAttributes (GetAttributesRequest) returns (GetAttributesResponse);
+  rpc UpdateMetadata (UpdateMetadataRequest) returns (Empty);
```
aaron-steinfeld, line 22: For a new API, please follow standard conventions
```suggestion
  rpc UpdateMetadata (UpdateMetadataRequest) returns (UpdateMetadataResponse);
```

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_service.proto
```diff
@@ -28,3 +29,14 @@ message GetAttributesRequest {
 message GetAttributesResponse {
   repeated AttributeMetadata attributes = 1;
 }
+
+message UpdateMetadataRequest {
+  AttributeIdentifier identifier = 1;
```
aaron-steinfeld, line 34: Why not just take an ID?

suresh-prakash:
Will add unit tests once the implementation is found good.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/builder/AttributeFilterBuilderImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+import static org.hypertrace.core.documentstore.expression.operators.RelationalOperator.EQ;
+
+import org.hypertrace.core.documentstore.expression.impl.ConstantExpression;
+import org.hypertrace.core.documentstore.expression.impl.IdentifierExpression;
+import org.hypertrace.core.documentstore.expression.impl.RelationalExpression;
+import org.hypertrace.core.documentstore.expression.type.FilterTypeExpression;
+
+public class AttributeFilterBuilderImpl implements AttributeFilterBuilder {
+  private static final String ID_SUB_DOC_PATH = "id";
```
aaron-steinfeld, line 11: guessing these are already defined elsewhere, would be good to avoid duplicating.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/delegate/AttributeUpdaterImpl.java
```diff
@@ -0,0 +1,93 @@
...
+  public UpdateMetadataResponse update(
+      final UpdateMetadataRequest request, final RequestContext context) throws IOException {
+    final String tenantId = AttributeMetadataValidator.validateContextAndGetTenantId(context);
+    validate(request);
+
+    final FilterTypeExpression filter = buildFilter(request, tenantId);
+    final Query query = Query.builder().setFilter(filter).build();
+    final List<SubDocumentUpdate> updates = buildUpdates(request);
+    final Optional<Document> docOptional =
+        collection.update(query, updates, DEFAULT_UPDATE_OPTIONS);
```
aaron-steinfeld, line 50: I might be missing it, but the tricky bit about updating here is the hierarchy which looks like it's missing. If I try to update a root-owned attribute, we should allow it, but we can't update the actual root attribute, otherwise we'd leak to other tenants. We need to create a clone that applies the update for the requesting tenant.

If I'm reading the code in this PR right, we're only supporting updates on attributes that are already defined in the requesting tenant's context. That may be sufficient for first cut, but at a minimum we should call it out.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/delegate/AttributeUpdaterImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        .map(converter::convert)
+        .map(Optional::orElseThrow)
+        .map(metadata -> UpdateMetadataResponse.newBuilder().setAttribute(metadata).build())
+        .orElseThrow();
+  }
+
+  private void validate(final UpdateMetadataRequest request) {
+    if (request.getAttributeId().isBlank()) {
+      throw Status.INVALID_ARGUMENT
+          .withDescription("Attribute id is must for updating")
```
aaron-steinfeld, line 69: nit: grammar
```suggestion
          .withDescription("Attribute id is required for updating")
```

aaron-steinfeld, line 69: nit: grammar

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/builder/AttributeUpdateBuilderImpl.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Override
+  public Optional<SubDocumentUpdate> buildUpdate(final Update update) {
+    final TypeCase typeCase = update.getTypeCase();
+    final Optional<SubDocumentUpdate> subDocumentUpdate =
+        Optional.ofNullable(UPDATE_PROVIDER_MAP.get(typeCase))
+            .map(provider -> provider.apply(update));
+
+    if (subDocumentUpdate.isEmpty()) {
```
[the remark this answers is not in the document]
suresh-prakash: Right. Updated.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/delegate/AttributeUpdaterImpl.java
suresh-prakash: Good point, Aaron. Will update it.

suresh-prakash: Updated the message `No attribute found with id: %s.` to `No custom attribute found with id: %s.`
I don't think we need to make an additional DB call to differentiate the actual non-existence of an attribute vs the existence of an attribute at the root level and supply the message accordingly. Do you agree?

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeMetadataValidator.java
```diff
@@ -103,6 +125,8 @@ private static void validate(AttributeMetadata attributeMetadata) {
         || attributeMetadata.getType().equals(AttributeType.TYPE_UNDEFINED)) {
       throw new IllegalArgumentException(String.format("Invalid attribute:%s", attributeMetadata));
     }
+
+    // TODO: Add length validation for all the string fields
```
suresh-prakash, line 129: Will do this in a follow-up PR since this has already grown sufficiently large.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/AttributeMetadataValidator.java
```diff
@@ -104,6 +106,15 @@ public static String validateContextAndGetTenantId(final RequestContext context)
                     .asRuntimeException());
   }
 
+  public static void validateMaxLength(final String string) {
+    if (string.length() > MAX_STRING_LENGTH) {
+      throw new IllegalArgumentException(
+          String.format(
```
aaron-steinfeld, line 112: status exceptions

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/constants/AttributeFieldPathConstants.java
```diff
@@ -0,0 +1,19 @@
...
+  public static final String FQN_PATH = "fqn";
+  public static final String SCOPE_PATH = "scope";
+  public static final String SCOPE_STRING_PATH = "scope_string";
+  public static final String KEY_PATH = "key";
+  public static final String INTERNAL_PATH = "internal";
+
+  public static final String TENANT_ID_PATH = "tenant_id";
+  public static final String SOURCE_METADATA_PATH = "metadata";
+
+  public static String sourceMetadataPathFor(final AttributeSource source) {
```
aaron-steinfeld, line 16: nit - not a constant, by definition 😉  - if it's only used in one spot, would leave it there

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/delegate/AttributeUpdaterImpl.java
```diff
@@ -80,14 +79,20 @@ private void validate(final UpdateMetadataRequest request) {
...
   private List<SubDocumentUpdate> buildUpdates(final UpdateMetadataRequest request) {
-    return request.getUpdatesList().stream()
-        .map(updateBuilder::buildUpdate)
-        .flatMap(Optional::stream)
-        .collect(toUnmodifiableList());
+    try {
+      return request.getUpdatesList().stream()
+          .map(updateBuilder::buildUpdate)
+          .collect(toUnmodifiableList());
+    } catch (final UnsupportedOperationException e) {
```
aaron-steinfeld, line 94: ah - I mentioned this elsewhere, but this works too. That said, why not just throw the status directly? Would be cleaner and no implicit contract.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/delegate/AttributeUpdaterImpl.java
aaron-steinfeld: Yes. If we're going to put in the effort to check existence, might as well add the support for updating attributes that aren't custom.

aaron-steinfeld (approved):

suresh-prakash:
Will address the review comments in a follow-up PR.

PR #156, opened 2022-12-08
Attribute update implementation fixes with string length validation for attribute creation

suresh-prakash:


attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/constants/AttributeFieldPathConstants.java
```diff
@@ -1,7 +1,5 @@
 package org.hypertrace.core.attribute.service.constants;
 
-import org.hypertrace.core.attribute.service.v1.AttributeSource;
-
 public class AttributeFieldPathConstants {
```
aaron-steinfeld, line 3: nit - could make this an interface now to prevent adding any methods

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/validator/StringLengthValidatorImpl.java
```diff
@@ -0,0 +1,63 @@
...
+public class StringLengthValidatorImpl implements StringLengthValidator {
+  private static final int MAX_STRING_LENGTH = 1000;
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+  private static final Printer PRINTER = JsonFormat.printer();
+
+  @Override
+  public void validate(final Message message) {
+    final Descriptor descriptor = message.getDescriptorForType();
+    try {
+      final JsonNode node = MAPPER.readTree(PRINTER.print(message));
```
aaron-steinfeld, line 27: Why are we converting a proto message to string then back to a generic json object? If we want this generic (which is questionable, given we only have one message here right?) then we should be working with at the proto level rather than json.

suresh-prakash: > If we want this generic (which is questionable, given we only have one message here right?)

Basically, we are trying to validate all the string values that are a part of the proto (even if they are nested inside other messages, one-ofs, or lists). Any string value introduced in the future would be automatically validated.

> we should be working with at the proto level rather than json

Right. Updated the implementation.

skjindal93 (approved):

aaron-steinfeld (approved):

PR #157, opened 2023-02-03
Rename admin port to http-admin

nikhil-rajesh:


helm/values.yaml
```diff
@@ -12,7 +12,9 @@ image:
 imagePullSecrets: {}
 
 containerPort: 9012
-containerHealthProbePort: 9013
+containerPortName: grpc-port
+containerAdminPort: 9013
+containerAdminPortName: http-admin
```
aaron-steinfeld, line 17: Context? I don't have any particular attachment to the names, but why do we need any port names here when we're binding by port numbers anyway?

nikhil-rajesh: Trying to have consistent port-names so that it'll be easier to parse from the yaml in the [debugging toolkit](https://github.com/Traceableai/debugging-toolkit) we are building.

PR #170, opened 2023-06-22
ci: remove pull_request_target

aaron-steinfeld:
## Description

Merge build and test workflows into same workflow and scheduled them. They had been separated to allow secret access in forks, but that's not really necessary - changed it to omit steps in forks that require secrets.


tim-mwangi (approved):

.github/workflows/build-and-test.yml
```diff
@@ -1,11 +1,39 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_READ_USER }}
+          password: ${{ secrets.DOCKERHUB_READ_TOKEN }}
+
+      - name: Build with Gradle
+        uses: hypertrace/github-actions/gradle@main
+        with:
+          args: assemble dockerBuildImages
+
+      - name: Run Trivy vulnerability scanner
```
ravisingal, line 32: should we add `if` condition here as well?

aaron-steinfeld: hmm let me go check inside the action - from the outside it appears safe to run on a fork

aaron-steinfeld: So the upload would presumably fail - perhaps change output mode for forks to log?

aaron-steinfeld: I had changed the local one here to 
```
output-mode: ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) && 'github' || 'log' }}
```
but if we can put that defaulting into the action it'd be cleaner.

aaron-steinfeld: backed my local change out in favor of the shared action change

ravisingal (approved):

PR #172, opened 2023-08-24
chore: poc of bom

aaron-steinfeld:
## Description

POC of shared bom and catalog for discussion.


attribute-projection-registry/build.gradle.kts
```diff
@@ -1,15 +1,42 @@
 plugins {
   `java-library`
   jacoco
-  id("org.hypertrace.jacoco-report-plugin")
-  id("org.hypertrace.publish-plugin")
+  alias(commonLibs.plugins.hypertrace.jacoco)
+  alias(commonLibs.plugins.hypertrace.publish)
+}
+
+configurations {
```
aaron-steinfeld, line 8: from here to line 32 would likely go into a plugin, which applies something similar to all java projects.

attribute-projection-registry/build.gradle.kts
```diff
@@ -1,15 +1,42 @@
...
+  runtimeClasspath {
+    resolutionStrategy.activateDependencyLocking()
+  }
+}
+
+dependencyLocking {
+  lockMode.set(LockMode.STRICT)
+}
+
+tasks.register("resolveAndLockAll") {
```
aaron-steinfeld, line 21: basically a utility to make it easy to generate the locks across all projects. Ref: https://docs.gradle.org/current/userguide/dependency_locking.html#ex-resolving-all-configurations

attribute-projection-registry/build.gradle.kts
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
 dependencies {
-  api(project(":attribute-service-api"))
-  implementation(project(":attribute-projection-functions"))
-  implementation("com.google.guava:guava:32.0.1-jre")
-  testImplementation("org.junit.jupiter:junit-jupiter:5.8.2")
+  api(projects.attributeServiceApi)
+  api(platform(commonLibs.hypertrace.bom))
```
aaron-steinfeld, line 36: On the fence if plugin should auto add bom to each project. Seems like it would make things easier.

settings.gradle.kts
```diff
@@ -8,6 +8,22 @@ pluginManagement {
   }
 }
 
+dependencyResolutionManagement {
```
aaron-steinfeld, line 11: This too would probably go in a plugin (probably replaces repo plugin as a settings plugin)

settings.gradle.kts
```diff
@@ -8,6 +8,22 @@ pluginManagement {
...
+  repositories {
+    mavenLocal()
+    mavenCentral()
+    maven("https://hypertrace.jfrog.io/artifactory/maven")
+  }
+  versionCatalogs {
+    create("commonLibs") {
+      from("org.hypertrace.bom:hypertrace-version-catalog:+")
```
aaron-steinfeld, line 19: so this would get the latest catalog and bom whenever locks are refreshed.

I was thinking we'd add a GHA job - probably scheduled with a manual option - that regenerates locks and raises a new PR. That means a bom change would percolate to all consuming repos on their own schedule, or can be manually forced. All builds continue to be reproducible due to lock.

attribute-projection-registry/build.gradle.kts
surajpuvvada: How does this work ? What does it ensure ?

aaron-steinfeld: This updates the lock files. The build will resolve all versions exactly from the lock files, meaning builds are reproducible (which they are anyways with static versions, but we're intentionally introducing dynamic versions here to support automatic upgrading).

So builds work normally, there's just the extra lock file (you'll see a couple in this PR). By running `gw resolveAndLockAll --write-locks` you can update those lock files. In a later comment - https://github.com/hypertrace/attribute-service/pull/172#discussion_r1304886875 - I discuss that I planning on adding an action that does this periodically/on trigger as well.

attribute-projection-registry/build.gradle.kts
mihirgt: If we are using version catalogs and doing dependency locking, what does the platform-plugin do additionally?

settings.gradle.kts
mihirgt: Which project is publishing this catalog? I checked here https://github.com/hypertrace/hypertrace-bom, did not find one.

attribute-projection-registry/build.gradle.kts
avinashkolluru: Does it mean once has to run this task locally to generate the new lock file if any dep changes are made? Otherwise, gradle would continue to use the existing ones in the file?

aaron-steinfeld: I'll definitely write up some documentation on this if we go forward with the locks, but in short:

- with any gradle command, you can append the `--write-locks` flag and it will update/create lock files for any dependencies it resolves.
- This task is there to make it easy to do that for all of the projects in a multi-project build at once (basically it forces resolution of all the configs). A built in way is just `gw dependencies --write-locks`, the issue with that is it works on one project at a time. Other commands, like `check` or `build` would also likely be sufficient for our case, but I figured this makes it a little more explicit and would handle projects using custom configurations that aren't resolved by those built in tasks too. Ref: https://docs.gradle.org/current/userguide/dependency_locking.html#lock_all_configurations_in_one_build_execution
- If you add or upgrade a dependency, then try to run the build without updating the lock, it will fail. If you _downgrade_ a dependency, that will not fail (basically only if conflict resolution doesn't match the lock, it's a failure).
image.

attribute-projection-registry/build.gradle.kts
aaron-steinfeld: Not sure if you read the link I sent out earlier, but it dives into this exact question. Including again here in case: https://docs.gradle.org/current/userguide/platforms.html#sub:platforms-vs-catalog , but let me see if I can do a better job with some examples. Apologies if this is repeating things that are already obvious, but I figure this will be a useful write up to share more broadly anyway so hopefully not wasted.

Let's take a project that previously declared:
```
implementation("com.google.guava:guava:32.0.1-jre")
```

A basic version catalog basically just gives type safe accessors. It just saves a user from typing that out which means the versions in that project will always be in sync.
```
implementation(libs.guava)
```

So far so good, but that's one repo. So now we can start publishing a version catalog, and we can do the same thing across repositories. I believe this is the experiment you were working with. But at this point, we've just kicked the problem down the road, because we have to declare the version of the catalog artifact itself that we want to use. Different repos may be using different catalogs. So that's where we introduced a dynamic version - always use the most recent version of catalog:
```
  versionCatalogs {
    create("commonLibs") {
      from("org.hypertrace.bom:hypertrace-version-catalog:+")
    }
  }
```

This introduces our next issue - repeatable builds. Having a build succeed today and then fail tomorrow because the version catalog changed with no source code changes sucks. So in come dependency locks. These basically materialize the resolution result and are checked in. There are equivalents in many languages (e.g. go, javascript/npm). So now, even if a new catalog is published, I have to opt in to pulling down that update.

Now what's the purpose of a dependency platform? It adds constraints on the resolution result even for dependencies you don't declare. So let's say we've got a second project, and this one has a dependency on grpc-api but is not directly using guava:
```
  implementation(libs.grpc.api) // "io.grpc:grpc-api:1.56.1"
```
Well grpc-api has a dependency on guava `"com.google.guava:guava:31.1-android"`, so this project would compile and run against that older version of guava - now we've got different projects compiling against different guava versions. When they're resolved together, conflict resolution has to select a single version for runtime (by default the newest one) - but that's a different version than the second project compiled against.

So when we add in the dependency platform on all projects, this aligns not just all of their first order dependencies, but their transitives too. Further - when we declare it as part of the API, that goes viral. 
```
  implementation(libs.grpc.api)
  api(platform(commonLibs.hypertrace.bom))
```
A project that depends on the above project will be depending on the same platform too and get the same benefit. Let's say some other service depends on this artifact, but it is using an older grpc version. It provides `grpc-netty:1.56.0`. Without the platform, it'd have misaligned versions using the newer grpc-api and the older grpc-netty, which can lead to runtime conflicts.  Because the platform _also_ declares the grpc-netty version, just by depending on this project we keep our versions all in sync.

I hope that helps and didn't just make things more confusing!

settings.gradle.kts
```diff
@@ -8,8 +8,11 @@ pluginManagement {
   }
 }
 
+enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
+
 plugins {
   id("org.hypertrace.version-settings") version "0.2.0"
+  id("org.hypertrace.dependency-settings") version "0.1.0-SNAPSHOT"
```
aaron-steinfeld, line 15: https://github.com/hypertrace/hypertrace-gradle-dependency-settings-plugin/pull/1

.github/actions/raise-lock-pr/action.yaml
```diff
@@ -0,0 +1,29 @@
+name: Update Gradle lockfiles and raise PR
```
aaron-steinfeld, line 1: I'll move this to shared repo in a separate PR. Going to test it here first.

aaron-steinfeld: (need to merge to test it since it's a manual/scheduled workflow)

caching-attribute-service-client/gradle.lockfile
```diff
@@ -0,0 +1,42 @@
+# This is a Gradle generated file for dependency locking.
```
tim-mwangi, line 1: Should this file be checked in?

tim-mwangi: So based on GHAs it's the whole point of the PR.

aaron-steinfeld: yep (also see line 3 😉 )

avinashkolluru (approved):

tim-mwangi (approved):

PR #186, opened 2023-09-19
fix: update config for new new doc store format

aaron-steinfeld:
## Description

New doc store config has some breaking changes (which should be fixed), but upgrading to use new format here.


tim-mwangi (approved):

attribute-service/src/main/resources/configs/common/application.conf
```diff
@@ -3,11 +3,17 @@ service.name = attribute-service
 service.admin.port = 9013
 document.store {
   dataStoreType = mongo
+  appName = attribute-service-local
```
suresh-prakash, line 6: Can we override the appName with the SERVICE_NAME environment variable (so that we can differentiate between a local run and a deployment)?

helm/templates/deployment.yaml
```diff
@@ -75,14 +75,22 @@ spec:
               containerPort: {{ .Values.containerHealthProbePort }}
               protocol: TCP
           env:
-            - name: SERVICE_NAME
```
suresh-prakash, line 78: We may need the service name if we want to override the app name.

attribute-service/src/main/resources/configs/common/application.conf
```diff
@@ -3,11 +3,17 @@ service.name = attribute-service
...
-    port = 27017
-    port = ${?MONGO_PORT}
+    database = default_db
+    endpoints = [
+      {
+        host = localhost
+        host = ${?MONGO_HOST} # provides a way to override the mongo_host via an environment variable
+        port = 27017
+        port = ${?MONGO_PORT}
+      }
```
suresh-prakash, line 15: Not consuming the mounted credentials (env. vars.)?

attribute-service/src/main/resources/configs/common/application.conf
aaron-steinfeld: > so that we can differentiate between a local run and a deployment

This goal is accomplished, just in a slightly different way. Instead of declaring that an env var can override this in the local conf, then in helm setting up the env var, we just directly override it in the helm conf using an existing value.

I preferred this approach for a couple reasons, nothing major though:
 - a few less lines of conf, and more standard with how other config overrides work
 - even in local runs, we often set service name as a jvm prop like below. So it seems confusing to rely on the _env var_ not being set, but the jvm prop to be set.
  ```kts
tasks.run<JavaExec> {
  jvmArgs = listOf("-Dservice.name=${project.name}")
}
```

helm/templates/deployment.yaml
aaron-steinfeld: Responded in first thread, can continue discussion there

attribute-service/src/main/resources/configs/common/application.conf
aaron-steinfeld: This was intended, but I could be convinced to change it.

My goal was basically to expose a way to use credentials without requiring them which would be a breaking change for anyone currently using attribute service. Basically keep the same default behavior while extending the flexibility of the config.  If we can map the default to env vars and it ignores them gracefully if env not set, then I'll go back and add em in - let me check that.

aaron-steinfeld: tested this, they're ignored if the env var isn't set so it works correctly. Added them in

attribute-service/src/main/resources/configs/common/application.conf
suresh-prakash: Makes sense

suresh-prakash (approved):

tim-mwangi (approved):

PR #192, opened 2023-10-11
Inject Mongo credentials optionally based on a helm flag

suresh-prakash:


nikhil-rajesh (approved):

attribute-service-factory/gradle.lockfile
```diff
@@ -76,7 +76,7 @@ org.eclipse.jetty:jetty-util-ajax:9.4.52.v20230823=runtimeClasspath
 org.eclipse.jetty:jetty-util:9.4.52.v20230823=runtimeClasspath
 org.hdrhistogram:HdrHistogram:2.1.12=runtimeClasspath
 org.hypertrace.bom:hypertrace-bom:0.2.6=compileClasspath,runtimeClasspath
-org.hypertrace.core.documentstore:document-store:0.7.37=compileClasspath,runtimeClasspath
```
aaron-steinfeld, line 79: Were these manual edits? If so, please update this in the bom repo first, then update the bom here.

aaron-steinfeld: Updated the bom, then generated the PR to consume that update. I've pointed that PR at this one, so feel free to merge it into your change - https://github.com/hypertrace/attribute-service/pull/193/

suresh-prakash: Thanks Aaron. For next time, can you guide with the steps that I should follow?

skjindal93 (approved):

attribute-service-factory/gradle.lockfile
aaron-steinfeld: So all dependencies for this service (and this will come to other services too, when I get time) are declared in https://github.com/hypertrace/hypertrace-bom/blob/main/gradle/libs.versions.toml . When publishing a new artifact, that's the spot to update for other services to consume.

Once a new bom is published, all repos consuming it have a weekly upgrade cadence that will pull in the latest changes, use them to update the gradle lockfiles, and publish a PR (like #193 - I just edited it to point it at this branch rather than main). If the weekly cadence isn't fast enough (e.g. here I just updated the bom and wanted to bring the change in), you can either manually trigger the scheduled run (as I did):
image
Or include the `--write-locks` flag with any gradle command. Further details in this comment (And generally in this PR): https://github.com/hypertrace/attribute-service/pull/172#discussion_r1305667616

So with this, we're trying to strike a balance of "always up to date", where dependencies automatically propagate, while still keeping builds reproducible. This will help make it easy to update vulnerabilities and propagate lib changes and by keeping all services in sync, we reduce the risk of conflicts when services are running co-resident in the same container/classloader. Note this also makes it even more important to avoid breaking changes. I'll write something up a bit more formally as we start rolling this out more widely.

PR #197, opened 2023-11-05
build: update bom

aaron-steinfeld:
## Description
Manual lock update and CI fix to support autobuild with locks


.github/workflows/codeql-analysis.yml
```diff
@@ -27,6 +27,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
aaron-steinfeld, line 31: The issue was hard to track down but eventually isolated it to the lack of history broke the local version calculation and led to the local build being mismatched with the locks (if the bom contains self references, since we recently added attribute service to bom).

skjindal93 (approved):

saxenakshitiz (approved):

ravisingal (approved):

PR #200, opened 2023-11-20
Add non rx based caching attribute client

kishansairam9:


caching-attribute-service-client/src/test/java/org/hypertrace/core/attribute/service/cachingclient/AttributeServiceCachedClientTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+import com.typesafe.config.ConfigFactory;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class AttributeServiceCachedClientTest extends CachingClientTestUtils {
```
aaron-steinfeld, line 15: May be irrelevant with the ask to split the client modules, but in general we don't want tests extending other tests (and here particularly, it certainly fails the is-a rule). If you want to share some common utils, that's fine, but inheritance isn't the right tool for that.

caching-attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/cachingclient/AttributeServiceCachedClient.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  public Optional<AttributeMetadata> get(
+      String tenantId, String attributeScope, String attributeKey) {
+    try {
+      return Optional.ofNullable(cache.getUnchecked(tenantId))
+          .map(table -> table.get(attributeScope, attributeKey));
+    } catch (Exception e) {
+      if (LOG_LIMITER.tryAcquire()) {
+        log.error("No attribute available for scope {} and key {}", attributeScope, attributeKey);
```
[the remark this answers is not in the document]
kishansairam9: Sorry my bad, fixing

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceCachedClient.java
```diff
@@ -30,102 +31,91 @@
...
-                      .map(scopeAndKey -> table.get(scopeAndKey.getScope(), scopeAndKey.getKey())));
-    } catch (Exception e) {
-      if (LOG_LIMITER.tryAcquire()) {
-        log.error("No attribute available for id {}", attributeId);
-      }
+      return getTableForRequestContext(requestContext)
+          .map(
+              table -> {
+                AttributeScopeAndKey scopeAndKey = scopeAndKeyLookup.getIfPresent(attributeId);
+                return table.get(scopeAndKey.getScope(), scopeAndKey.getKey());
```
aaron-steinfeld, line 92: Now here we could throw a NPE - I would suggest fixing that, should behave identically to the scope/key version

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceCachedClient.java
```diff
@@ -30,102 +31,91 @@
...
+      @Nonnull RequestContext requestContext, @Nonnull String attributeScope)
+      throws ExecutionException {
+    return getTableForRequestContext(requestContext)
+        .map(table -> List.copyOf(table.row(attributeScope).values()))
+        .orElse(Collections.emptyList());
+  }
+
+  private Optional<Table<String, String, AttributeMetadata>> getTableForRequestContext(
+      RequestContext requestContext) throws ExecutionException {
+    return Optional.of(
```
aaron-steinfeld, line 110: There's no point in returning an optional here if it can never be empty

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceCachedClient.java
```diff
@@ -30,102 +31,91 @@
...
+        cache.get(
+            requestContext
+                .getTenantId()
```
aaron-steinfeld, line 113: We want the request context in the key, not using the tenant ID directly. If you don't want to use the context key generated from the request context you can build your own, but either way if I call client.getById and the client needs to make it a network call, it should be using the exact request context I provided to it.

attribute-service-client/src/main/java/org/hypertrace/core/attribute/service/client/AttributeServiceCachedClient.java
```diff
@@ -30,102 +31,91 @@
...
+      throws ExecutionException {
     try {
-      return Optional.ofNullable(cache.getUnchecked(tenantId))
+      return getTableForRequestContext(requestContext)
           .map(table -> table.get(attributeScope, attributeKey));
-    } catch (Exception e) {
-      if (LOG_LIMITER.tryAcquire()) {
-        log.error("No attribute available for scope {} and key {}", attributeScope, attributeKey);
-      }
+    } catch (NullPointerException e) {
```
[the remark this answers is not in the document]
kishansairam9: I misread the docs of optional.map 🤕. It throws NPE when mapper is null, fixing

PR #209, opened 2024-01-19
Upgrade gradle to 8.5

patelraj0602:

PR #214, opened 2024-01-24
Gradle update

patelraj0602:


kishansairam9 (approved):

aaron-steinfeld (changes requested):
Dupe of #209 , please close this one (or that one, if you want) and address the comments on that (both expired suppressions are not needed, so please remove them rather than extending their expirations - you removed one and extended one here)

aaron-steinfeld (approved):

PR #233, opened 2024-07-16
Add new field "description" in attribute_metadata.proto

siddhant2001:
Adding support for description for an attribute in the AttributeMetadata class. The description is expected to be used for documentation down the line.

Default value for the new field is an empty string.

### DIFF
--- a/attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
+++ b/attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
@@ -103,6 +103,7 @@ message AttributeMetadata {
   string scope_string = 17;
   bool internal = 18;
   bool custom = 19;
+  string description = 20;
 }
 
 message AttributeSourceMetadata {
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
@@ -53,9 +53,6 @@ public class AttributeMetadataModel implements Document {
   @JsonProperty(value = "display_name")
   private String displayName;
 
-  @JsonProperty(value = "description")
-  private String description;
-
   @JsonProperty(value = "scope_string")
   private String scopeString;
 
@@ -79,6 +76,9 @@ public class AttributeMetadataModel implements Document {
   private AttributeDefinition definition = AttributeDefinition.getDefaultInstance();
 
   private boolean internal;
+
+  @JsonProperty(value = "description")
+  private String description = "";
 
   protected AttributeMetadataModel() {}
 
@@ -394,8 +394,8 @@ public class AttributeMetadataModel implements Document {
         && Objects.equals(metadata, that.metadata)
         && Objects.equals(definition, that.definition)
         && Objects.equals(scopeString, that.scopeString)
-        && description.equals(that.description)
-        && Objects.equals(internal, that.internal);
+        && Objects.equals(internal, that.internal)
+        && Objects.equals(description, that.description);
   }
 
   @Override
@@ -417,8 +417,8 @@ public class AttributeMetadataModel implements Document {
         metadata,
         definition,
         scopeString,
-        description,
-        internal);
+        internal,
+        description);
   }
 
   private static class ProtobufMessageSerializer extends JsonSerializer<Message> {
--- a/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
+++ b/attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
@@ -53,6 +53,9 @@ public class AttributeMetadataModel implements Document {
   @JsonProperty(value = "display_name")
   private String displayName;
 
+  @JsonProperty(value = "description")
+  private String description;
+
   @JsonProperty(value = "scope_string")
   private String scopeString;
 
@@ -97,6 +100,7 @@ public class AttributeMetadataModel implements Document {
     attributeMetadataModel.setSources(attributeMetadata.getSourcesList());
     attributeMetadataModel.setGroupable(attributeMetadata.getGroupable());
     attributeMetadataModel.setDefinition(attributeMetadata.getDefinition());
+    attributeMetadataModel.setDescription(attributeMetadata.getDescription());
     attributeMetadataModel.setMetadata(
         attributeMetadata.getMetadataMap().entrySet().stream()
             .collect(
@@ -146,6 +150,14 @@ public class AttributeMetadataModel implements Document {
 
   public void setDisplayName(String displayName) {
     this.displayName = displayName;
+  }
+
+  public void setDescription(String description) {
+    this.description = description;
+  }
+
+  public String getDescription() {
+    return description;
   }
 
   public String getScopeString() {
@@ -294,6 +306,9 @@ public class AttributeMetadataModel implements Document {
     if (!definition.equals(AttributeDefinition.getDefaultInstance())) {
       builder.setDefinition(definition);
     }
+    if (description != null) {
+      builder.setDescription(description);
+    }
 
     return builder;
   }
@@ -349,6 +364,8 @@ public class AttributeMetadataModel implements Document {
         + metadata
         + ", internal="
         + internal
+        + ", description="
+        + description
         + '}';
   }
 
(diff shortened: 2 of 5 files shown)


attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -53,6 +53,9 @@ public class AttributeMetadataModel implements Document {
   @JsonProperty(value = "display_name")
   private String displayName;
 
+  @JsonProperty(value = "description")
+  private String description;
```
kotharironak, line 57: The existing documents will not have this property. We need to check if it defaults to null in that case.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -377,6 +394,7 @@ public boolean equals(Object o) {
         && Objects.equals(metadata, that.metadata)
         && Objects.equals(definition, that.definition)
         && Objects.equals(scopeString, that.scopeString)
+        && description.equals(that.description)
```
kotharironak, line 397: nit: new addition at the end - can we move this condition after `&& Objects.equals(internal, that.internal)`

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -399,6 +417,7 @@ public int hashCode() {
         metadata,
         definition,
         scopeString,
+        description,
```
kotharironak, line 420: Same as above.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
kotharironak, line 397: Change: description.equals(that.description) -> Objects.equals(description, that.description)

Note: If `description` is null, the check `description.equals(that.description)` will cause an exception.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -53,6 +53,9 @@ public class AttributeMetadataModel implements Document {
   @JsonProperty(value = "display_name")
   private String displayName;
 
+  @JsonProperty(value = "description")
```
kotharironak, line 56: nit (up to you): we can also move this definition after `private boolean internal;` so that it is the last addition in the object.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
kotharironak: This should be fine as `ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)`, but lets see if we have a test for it.

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
```diff
@@ -504,6 +507,10 @@ private Document createMockDocument(
         .append(kind.name())
         .append("\",\"tenant_id\":\"")
         .append(tenantId)
+        .append("\",\"description\":\"")
+        .append(scope.name())
```
kotharironak, line 511: We are copying the `scope.name()` as the description. 
Can we keep the existing tests as they are and add an additional test for the description? Since the `description` string is optional, the existing tests should work fine, right?

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModelTest.java
```diff
@@ -255,10 +258,12 @@ public void testAttributeDefinitionBackwardsCompatibility() throws IOException {
             + "\"id\":\"EVENT.key\","
             + "\"value_kind\":\"TYPE_STRING\","
             + "\"display_name\":\"Display\","
+            + "\"description\":\"\","
```
kotharironak, line 261: If it's without definitionAndDescription, we don't need an empty `""` string, right? Are we missing something?

Can you add an explicit test `testDescriptionBackwardsCompatibility`?

kotharironak: >If it's without definitionAndDescription, we don't need an empty "" string, right? Are we missing something?

I think this is because the default string value is an empty string in protobuf, and this test has two parts - one from `fromJson`, and the second from `fromDTO`. Got it.

kotharironak: Can you add an explicit test testDescriptionBackwardsCompatibility?

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
siddhant2001: Done.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
siddhant2001: Done.

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
siddhant2001: Done both.

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/AttributeServiceImplTest.java
siddhant2001: You are right, I've undone any changes I made to this file. Do we need separate tests for description apart from the ones added in AttributeMetadataModelTest?

attribute-service-impl/src/test/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModelTest.java
siddhant2001: I've reverted the `testAttributeDefinitionBackwardsCompatibility` to it's original state and added new tests in `testAttributeDescriptionBackwardsCompatibility`

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
```diff
@@ -103,6 +103,7 @@ message AttributeMetadata {
   string scope_string = 17;
   bool internal = 18;
   bool custom = 19;
+  string description = 20;
```
skjindal93, line 106:
```suggestion
  optional string description = 20;
```

siddhant2001: This change may not be required. Not setting the description value will not throw an error. It will just be set to it's default value, i.e., an empty string. 
Is there any other reason for having it as optional?

skjindal93: I was mostly thinking from client’s perspective, if they should be handling empty strings (they would anyways have to I guess, even if the field is set), or testing for the presence of description field

We should be good without optional too

attribute-service-impl/src/main/java/org/hypertrace/core/attribute/service/model/AttributeMetadataModel.java
```diff
@@ -77,6 +77,9 @@ public class AttributeMetadataModel implements Document {
 
   private boolean internal;
 
+  @JsonProperty(value = "description")
+  private String description = "";
```
kotharironak, line 81:
```suggestion
  private String description;
```

attribute-service-api/src/main/proto/org/hypertrace/core/attribute/service/v1/attribute_metadata.proto
siddhant2001: In both the cases a) old record without description and b) new record created without a description, the object returned by graphql will have an empty string for description.

kotharironak (approved):

kotharironak (approved):

